Issue #430, opened 2026-04-21
[Bug] [CLI Wizard] Custom service auth method options are incomplete compared with catalog / Web UI support


wanghuan-520:
### Environment

- Hosted API: `version = "0.2.0"`, `commit = "dfbe8fd526b8"`
- CLI: `NyxID CLI v0.2.0 (542cd7465b87-dirty)`
- Test flow: `nyxid service add` browser wizard

### Steps to Reproduce

1. Run `NYXID_WIZARD_NO_OPEN=1 nyxid service add`.
2. Open the local wizard URL printed by the CLI in Chrome.
3. First select the existing catalog entry `Discord Bot API`, click `Next`, and confirm that this path explains it will inject `Authorization: Bot <token>`.
4. Complete creation with a dummy token, then inspect the created service and confirm it is saved with `auth_method = bot_bearer`.
5. Delete the temporary service.
6. Open the wizard again and select `Custom / self-hosted...`.
7. Click `Next`.
8. Open the `Auth method` dropdown.

### Actual Result

The existing catalog entry `Discord Bot API` can be created successfully and is saved as:

- `endpoint_url = https://discord.com/api/v10`
- `auth_method = bot_bearer`
- `auth_key_name = Authorization`
- `credential_type = api_key`

However, the `Auth method` dropdown in the `Custom / self-hosted...` form only shows:

- `bearer`
- `header`
- `query`
- `basic`
- `none`

`bot_bearer` and `body` are not available in the Custom service browser wizard.

### Expected Result

The Custom service wizard should expose the complete auth method set supported by the custom service flow, or at least stay aligned with capabilities already available in catalog / Web UI flows, including:

- `bot_bearer`: for downstream services that require `Authorization: Bot <credential>`
- `body`: for downstream services that require credential injection into the request body

If the Custom service browser wizard intentionally does not support these methods yet, the UI and CLI help should make that limitation explicit so users do not assume all custom auth methods can be created through the wizard.

### Impact

Users can use the catalog-provided `Discord Bot API`, but they cannot create an equivalent custom `Authorization: Bot <token>` service through `Custom / self-hosted...`. They also cannot create custom services that require body-injected credentials.

This makes catalog service capabilities and custom service capabilities inconsistent. If users want to customize the endpoint or slug, or create a service similar to `Discord Bot API`, they must fall back to another CLI/API path. It also prevents the custom service auth method injection matrix from passing.

ctkm-aelf:
[...]

Issue #449, opened 2026-04-22
[Bug] [CLI Wizard] Device-code wizard disconnects when the user switches to the provider authorization page


wanghuan-520:
### Environment

- Hosted API: `version = "0.2.0"`, `commit = "c14d691349a4"`
- CLI: `NyxID CLI v0.2.0 (d950090fdeec-dirty)`
- Flow: `nyxid service add` browser wizard, device-code provider

### Steps to Reproduce

1. Run the browser wizard for a device-code provider:

```bash
NYXID_WIZARD_NO_OPEN=1 nyxid service add llm-openai-codex \
  --label "QA AK088 Device Code <timestamp>" \
  --base-url https://nyx-api.chrono-ai.fun
```

2. Open the local wizard URL printed by the CLI in Chrome.
3. Confirm that the wizard displays a device code and provider verification URL:

```text
Visit
https://auth.openai.com/codex/device
Waiting for authorization (polling every 5s)...
```

4. Open the provider verification URL in another tab, or switch away from the wizard tab to complete authorization.
5. Observe the wizard tab and CLI output.

### Actual Result

The wizard tab can enter a disconnected state while the user is trying to complete provider authorization:

```text
Wizard disconnected
The nyxid CLI is no longer running. Close this tab and re-run nyxid service add.
Your terminal has any partial state that was saved.
```

The CLI exits with:

```text
Browser stopped responding (tab closed?) — cancelling.
Wizard cancelled. No service was created.
```

In the retest, the temporary service was left in `pending_auth` until it was manually deleted:

```json
{
  "label": "QA AK088 Device Code CDP 202604221555",
  "slug": "llm-openai-codex-2",
  "status": "pending_auth",
  "credential_type": "oauth2"
}
```

This appears to happen because the wizard page stops sending heartbeat when it becomes hidden / backgrounded.

However, switching away from the wizard page is a normal user action during device-code authorization. The user must leave the wizard page to open the provider verification URL, enter the code, and complete authorization.

### Expected Result

Switching away from the wizard tab to complete provider authorization should not be treated as a closed / dead wizard tab.

The device-code flow should allow the user enough time to:

- copy or open the provider verification URL
- switch tabs or windows
- enter the device code
- approve the provider authorization
- return to the wizard, or let the wizard poll complete automatically

While device-code polling is active, the CLI should keep the wizard session alive, or use a liveness strategy that does not depend on the wizard tab remaining visible.

[...]

PR #552, opened 2026-04-28
fix(cli): widen wizard heartbeat first-ping window for NYXID_WIZARD_NO_OPEN

ctkm-aelf:
## Summary

`NYXID_WIZARD_NO_OPEN=1` prints the wizard URL but skips `open::that(url)` — the user pastes the URL into a browser by hand, so the first heartbeat ping arrives on human time. The watchdog's no-first-ping deadline was hardcoded to `HEARTBEAT_STARTUP_GRACE + HEARTBEAT_DEAD_AFTER = 5 s + 4 s = 9 s`, which races with manual copy-paste and kills the wizard before the user can load the URL.

Practical impact: the env var documented in `server.rs:1465` as "used by automated validation and CI smoke tests" is unusable for the manual-paste scenarios it implies. Reproduced today during prod testing — the wizard prints `Browser stopped responding (tab closed?) — cancelling.` ~9 s after `service add` even when the user is actively trying to copy the URL.

## Fix

Widen the no-first-ping deadline to **20 s** specifically when `NYXID_WIZARD_NO_OPEN` is set. Once the browser pings even once, `HEARTBEAT_DEAD_AFTER` (4 s) takes over as before — a tab that opens then dies is still cleaned up at the normal cadence. Default behavior is unchanged when the env var is unset.

### Diff shape

- **New constant** `HEARTBEAT_FIRST_PING_DEADLINE_NO_OPEN = 20 s` next to the other heartbeat constants in `cli/src/wizard/server.rs`.
- **Watchdog** reads `NYXID_WIZARD_NO_OPEN` once at startup and binds `first_ping_deadline` to either the new 20 s value or the existing `HEARTBEAT_STARTUP_GRACE + dead_after` sum.
- **Match arm** for `last == None` now compares against `first_ping_deadline`. The `Some(t)` arm is untouched — once a real browser is live, `dead_after` semantics are unchanged.

Single-file, ~20 lines added.

## Why no issue

Surfaced inline during issue #430 verification. Tiny scoped fix, single regression, no design debate — opening directly per Calvin's call.

## Test plan

- [x] `cargo build -p nyxid-cli` — clean
- [x] `cargo test -p nyxid-cli` — 234 + 1 pass (no change in count)
- [x] `cargo clippy -p nyxid-cli --all-targets -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] Pre-commit hook (`cargo fmt` + `cargo clippy`) green
- [ ] Manual smoke (post-merge or locally): `NYXID_WIZARD_NO_OPEN=1 nyxid service add` — copy-paste URL into a browser within ~15 s, wizard stays alive (was: cancelled at ~9 s).
- [ ] Manual regression (post-merge or locally): `nyxid service add` (env var unset) — auto-launch browser still works, default 9 s deadline unchanged.

## Out of scope

- Doc/help text mentioning the new 20 s window — current docs don't quantify the timeout, so nothing to update.
- Changing the `Some(t)` `dead_after` window — intentional; an active browser going silent for >4 s should still be treated as a closed tab.

### DIFF
--- a/cli/src/wizard/server.rs
+++ b/cli/src/wizard/server.rs
@@ -116,6 +116,16 @@ const HEARTBEAT_DEAD_AFTER_ROTATION: Duration = Duration::from_secs(30);
 /// Grace period at startup before we start enforcing the heartbeat dead
 /// line. Lets the browser actually load the page.
 const HEARTBEAT_STARTUP_GRACE: Duration = Duration::from_secs(5);
+/// First-heartbeat deadline when `NYXID_WIZARD_NO_OPEN` is set. With
+/// auto-open the browser is loading by the time the watchdog starts,
+/// so the default `HEARTBEAT_STARTUP_GRACE + HEARTBEAT_DEAD_AFTER`
+/// (~9 s) is enough. Without auto-open the user pastes the URL by
+/// hand, so the first ping arrives on human time — 9 s reliably kills
+/// the wizard mid copy-paste. 20 s gives a comfortable manual window
+/// without giving up the eventual cleanup of a wizard nobody loaded.
+/// Once the first ping lands, `HEARTBEAT_DEAD_AFTER` takes over so a
+/// truly-dead tab is still detected at the normal cadence.
+const HEARTBEAT_FIRST_PING_DEADLINE_NO_OPEN: Duration = Duration::from_secs(20);
 /// How often the CLI checks the last-heartbeat timestamp. 500 ms is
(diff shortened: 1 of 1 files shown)


ctkm-aelf:
Closing — this fix will be folded into the broader heartbeat grace-period work for #449 rather than landing as a standalone tweak.

The diff (~20 lines, single file) remains on branch `fix/wizard-no-open-heartbeat-grace` if it's useful as a starting point for the #449 PR.

Issue #651, opened 2026-05-07
[Bug] [CLI Wizard OAuth] Browser/CLI account mismatch causes generic `Session mismatch` and leaves wizard waiting


wanghuan-520:
When starting an OAuth provider connection from the CLI Wizard, users can hit an unclear account-mismatch failure if the CLI is logged in as one NyxID account while the browser is logged in as another. After provider-side authorization, the NyxID callback fails with a generic error instead of explaining the account mismatch:

```text
Connection Failed
Session mismatch
```

The local CLI Wizard remains stuck on:

```text
Waiting for provider authorization...
```

This is confusing because the user may have completed the provider authorization successfully, but NyxID does not explain that the CLI account and browser account are different.

### Reproduction

1. Log in to the CLI as account A:

```bash
nyxid status
```

Example observed account:

```text
Account: account-a@example.com
```

2. In the browser, log in to NyxID Dashboard as account B.

Example observed browser account:

```text
account-b@example.com
```

3. Start a CLI Wizard OAuth service add flow, for example Lark:

```bash
NYXID_WIZARD_NO_OPEN=1 nyxid service add --label "QA-OAUTH-MISMATCH-$(date +%s)" --base-url https://nyx-api.chrono-ai.fun
```

4. Open the local wizard URL.
5. Select an OAuth provider such as `Lark API (User OAuth)`.
6. Complete provider-side authorization in the browser.
7. Observe the NyxID callback page and the local wizard page.

### Actual

The NyxID callback page shows:

```text
Connection Failed
Session mismatch
```

The local CLI Wizard continues to show:

```text
Waiting for provider authorization...
```

The user is not told that the OAuth flow was started for CLI account A while the browser was signed in as account B.

### Expected

NyxID should handle this account mismatch explicitly and make the error actionable.

Expected callback page behavior:

```text
This OAuth flow was started for <cli-account>, but this browser is signed in as <browser-account>. Please switch accounts and retry.
```

Expected local CLI Wizard behavior:

- It should stop waiting.
- It should transition to a clear failed state.
- The CLI terminal should also exit or print an actionable account-mismatch message.

Optional UX improvements:

- Show the CLI account in the local wizard before opening provider sign-in.
- Warn users to use the same NyxID account in the browser.
- If technically possible, avoid relying on the browser's current NyxID session for CLI-initiated OAuth callbacks.

### Scope

[...]

PR #789, opened 2026-05-18
fix(oauth): surface account-mismatch error to browser and CLI (#695)

### LINKED ISSUE #695: [Bug] [CLI Wizard OAuth] Account-mismatch callback redirects to the browser account's AI Services page instead of showing the mismatch error immediately
During CLI Wizard OAuth account-mismatch regression testing, the backend now stores an actionable mismatch error on the failed credential, but the browser does not show that error when the OAuth callback fails.

The flow was started from the CLI under one NyxID account, while the browser was signed in to a different NyxID account. After completing provider-side OAuth authorization, the browser redirected to the signed-in browser account's `AI Services` list page instead of showing an account-mismatch failure page.

The actionable error was only visible later after switching back to the CLI account and opening the failed credential:

```text
This OAuth flow was started by h***@gmail.com, but this browser is signed in as s***@gmail.com. Sign out of NyxID in this browser (or switch to the CLI account) and retry.
```

This means the backend failure state and error message exist, but they are not surfaced at the time and place where the user needs them.

### Reproduction

1. Log in to the NyxID CLI as account A.
2. Log in to the NyxID Dashboard in the browser as account B.
3. Start an OAuth service add flow from the CLI Wizard, for example Lark:

```bash
NYXID_WIZARD_NO_OPEN=1 nyxid service add api-lark \
  --label "QA-I36-MISMATCH-$(date +%s)" \
  --base-url https://nyx-api.chrono-ai.fun
```

4. Open the local wizard URL in the browser session that is signed in as account B.
5. Continue to provider sign-in and approve the provider OAuth authorization.
6. Observe the callback result in the browser.
7. Switch back to account A and inspect the failed OAuth credential.

### Actual

After provider authorization, the browser redirects to account B's `AI Services` list page.

No account-mismatch error is shown during the callback flow.

The actionable mismatch message is only visible after switching to account A and opening the failed credential.

### Expected

The OAuth callback should immediately show a failure page with the actionable account-mismatch message, for example:

```text
Connection Failed

This OAuth flow was started by h***@gmail.com, but this browser is signed in as s***@gmail.com. Sign out of NyxID in this browser (or switch to the CLI account) and retry.
```

The local CLI Wizard should also transition to a clear failed terminal state and show the same actionable message, instead of requiring the user to discover the error later from the credential detail page owned by the initiating account.

### Impact

- Users may think OAuth succeeded because the browser redirects to the Dashboard.
- The real failure reason is hidden on a failed credential owned by the CLI account.
- Multi-account users still cannot understand what happened at the time of failure.
- From the user's perspective, Issue #651 is only partially fixed.

ctkm-aelf:
## Summary

Closes #695. Minimal fix that surfaces the account-mismatch error message the backend already computes (#651), so it actually reaches the user.

## What changes (+20 / −1)

- **`frontend/src/router.tsx`** — `/providers` parent route's `beforeLoad` redirects non-admin users to `/keys` before `/providers/callback` can render, discarding `?status=error&message=…`. Change: allow `/providers/callback` past the guard so any signed-in user lands on the callback page and sees the message.
- **`cli/src/commands/service.rs`** — `wait_for_authorized_key` collapsed every non-`active` status into `"OAuth flow finished with unexpected key status: <other>"`, dropping `error_message`. Change: add a `"failed"` arm that bails with `error_message` verbatim (fallback `"OAuth authorization failed"` if empty).

No backend changes — `error_message` is already written by #651.

## Scope and trigger conditions

This is an **edge case**. The bug only fires when **all four** conditions align:

1. User has two NyxID accounts.
2. Both reachable from the same browser session.
3. CLI signed in as account A; the same browser's dashboard tab is signed in as account B.
4. OAuth provider has server-side `oauth_client_id` / `oauth_client_secret` configured in the catalog.

Mode B (pair mode) is already guarded at `backend/src/services/cli_pairing_service.rs:295` — cross-user pairing claims return 404. Only Mode A (single-machine wizard) can hit this, and only with the cross-account session collision above. For typical single-account users this code path is unreachable.

## Why ship it anyway

- The diff is trivial (20 lines, no backend, no schema, no new deps).
- #651 already paid the cost of computing the actionable message; throwing it away by gating it on an admin guard the FE shouldn't have on this route is silly.
- When the rare collision does fire, current behavior is actively misleading (browser appears to succeed → AI Services list; CLI prints a confusing generic error). Bad failure modes erode trust even when rare.

## Follow-ups (separate PRs, not blocking this one)

- Fail fast at wizard start — compare `oauth_state.user_id` to the browser session before redirecting to the provider, eliminating the wasted round-trip.
- Audit `/providers` parent guard — confirm `/providers/callback` was unintentionally caught and nothing else is wrongly gated.
- Reconcile Mode A / Mode B — Mode B guards cross-account claims; Mode A doesn't. Pick a single stance.

## Test plan

- [ ] Frontend type-check: `cd frontend && npm run build` → exit 0.
- [ ] CLI build: `cargo build -p nyxid-cli` → exit 0.
- [ ] Backend mismatch tests (already in place, covers the data this PR depends on): `cargo test -p nyxid-server -- generic_oauth_callback_session_mismatch`.
- [ ] Reproduce the full flow against an environment with a server-side-OAuth-configured provider (staging): CLI as A, dashboard as B, run `nyxid service add <slug> --oauth`. Expect callback page to render the masked-email mismatch message, and CLI to exit non-zero with the same message verbatim.
- [ ] Regression: as a non-admin, navigate to `/providers` (no `/callback`) — still redirects to `/keys`. Guard relaxation is callback-only.

### DIFF
--- a/cli/src/commands/service.rs
+++ b/cli/src/commands/service.rs
@@ -1595,6 +1595,18 @@ async fn wait_for_authorized_key(api: &mut ApiClient, key_id: &str) -> Result<Va
                 eprintln!();
                 return Ok(key);
             }
+            "failed" => {
+                // The backend stores an actionable reason on the credential
+                // (issue #651, e.g. an account-mismatch message). Surface it
+                // verbatim instead of swallowing it under a generic status
+                // string -- issue #695.
+                eprintln!();
+                let msg = key["error_message"]
+                    .as_str()
+                    .filter(|s| !s.is_empty())
+                    .unwrap_or("OAuth authorization failed");
+                bail!("{msg}");
+            }
             other => {
                 eprintln!();
                 bail!("OAuth flow finished with unexpected key status: {other}");
--- a/frontend/src/router.tsx
+++ b/frontend/src/router.tsx
@@ -334,7 +334,14 @@ const connectionsRedirectRoute = createRoute({
 const providersRedirectRoute = createRoute({
   path: "/providers",
   getParentRoute: () => dashboardLayout,
-  beforeLoad: () => {
+  beforeLoad: ({ location }) => {
+    // The backend's OAuth callback redirects every authenticated user here
+    // with ?status=error&message=... on account mismatch (issue #695).
+    // ProvidersCallbackPage must render for non-admins too; otherwise the
+    // actionable error stored by issue #651 never reaches the user.
+    if (location.pathname === "/providers/callback") {
+      return;
+    }
     const { user } = useAuthStore.getState();
     if (user?.is_admin) {
       // Admin users can still access the providers management pages


ctkm-aelf:
Closing as not-shipping. Repro requires four conditions to align simultaneously (two NyxID accounts, both reachable from one browser, CLI + dashboard signed in as different ones, provider with server-side OAuth client creds). Mode B (pairing) is already guarded at backend/src/services/cli_pairing_service.rs:295; only Mode A can theoretically hit this and only in the specific cross-account session collision above. Filing follow-ups for fail-fast preflight + parent-guard audit. Closing #695 too.

PR #967, opened 2026-06-17
Add compute pool scheduler MVP

AlyciaBHZ:
## Summary
- add compute pool, task, and worker models/services/routes with worker-pull scheduling
- add nyxid compute CLI commands for pool management, task submit/result/cancel/status, and trusted local worker execution
- document security boundary, manual QA, and task retention config; include a local mock backend for pre-GPU testing

## Safety boundary
- workers pull tasks with a rotatable nyx_cwk_ token; worker tokens are stored as hashes
- NyxID never stores local endpoint URLs or local backend bearer tokens
- NyxID does not SSH into hosts or run arbitrary commands; hosts execute only against their configured local endpoint
- compute task audit events are metadata-only and do not include prompt/output bodies

## Tests
- cargo fmt --all
- cargo test -p nyxid compute_
- cargo check -p nyxid
- cargo check -p nyxid-cli

## Notes
- This is PR1: compute pool MVP. It intentionally does not extract a shared Oracle/compute relay scheduler core yet.
- Follow-up PR2 should introduce shared relay/scheduler primitives used by compute and future relay-style features without merging their business modules.

PR #969, opened 2026-06-17
Add standalone compute pool service integration

AlyciaBHZ:
## Summary
- add a standalone compute-pool-service integration instead of a NyxID core-native compute API
- provide a reference worker-pull queue service, trusted worker bridge, and OpenAPI spec for the consumer API
- document how to register the service through existing NyxID service + Credential Node flows

## Architecture
NyxID remains the control plane: org/agent auth, service governance, credential injection, node routing, proxying, and audit metadata. Compute-specific task state, worker tokens, local model endpoints, and GPU worker protocol stay in the external service.

This PR is not a NyxID service-pool framework. Cross-service counting, quotas, metering, and load balancing should be handled by a future generic NyxID service-pool design rather than by a compute-specific core API. The compute service exposes /v1/status as a capacity signal that such a layer could use later.

## Security boundary
- no NyxID backend /api/v1/compute routes or compute collections
- no NyxID org model changes
- NyxID does not SSH into worker hosts or execute shell commands
- worker-local endpoint URLs and backend tokens are not stored in NyxID
- with Credential Node routing, the compute service API token can stay on the node host

## Tests
- node --check integrations/compute-pool-service/server.mjs
- node --check integrations/compute-pool-service/worker.mjs

## Notes
This replaces the closed prototype PR #967 direction. The old PR proved the worker-pull UX and safety boundary, but this PR keeps compute as a NyxID-managed external service rather than adding compute as a core backend feature.

### DIFF
--- a/integrations/compute-condor-adapter/README.md
+++ b/integrations/compute-condor-adapter/README.md
@@ -1,0 +1,123 @@
+# Compute HTCondor Adapter
+
+This directory documents the production-oriented adapter shape for exposing an
+HTCondor pool through NyxID as an external service. It intentionally does not
+add NyxID backend routes, MongoDB models, org changes, or a fake partial
+HTCondor implementation.
+
+NyxID remains the control plane: identity, org membership, agent API keys,
+credential brokering, node routing, proxying, audit, and service governance.
+HTCondor remains the scheduler: execute points, submit queue, resource
+matching, preemption, retries, and idle-machine policy. The adapter is the
+thin data-plane service between them.
+
+## Architecture
+
+```text
+agent / org user
+  -> NyxID proxy / service governance
+  -> optional NyxID Credential Node
+  -> compute-condor-adapter
+  -> HTCondor submit host
+  -> HTCondor execute points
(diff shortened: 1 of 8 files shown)


AlyciaBHZ:
Addressed the two robustness notes that are most relevant before real concurrent traffic:

- Serialized JSON store writes through a process-local save queue and unique per-write tmp file, so concurrent requests in one process no longer race on the same tmp path.
- Made worker ack failures retryable. Transient /worker/ack errors no longer immediately fail the task; after --max-ack-failures consecutive failures, the worker aborts the local request and reports failure so compute is not orphaned.

Also tightened the docs/spec items:

- Documented the lease/ack relationship and the single consumer token task-visibility boundary.
- Updated OpenAPI server URL to 127.0.0.1 and added the 409 task_terminal cancel response.
- Re-ran node --check for server.mjs and worker.mjs, plus git diff --check.

Leaving functional state-machine tests and durable storage as follow-up work for when this reference integration grows beyond the smoke-test JSON backend.

PR #979, opened 2026-06-19
feat(oracle): pin follow-up tasks to the owning worker (session→worker affinity)

### LINKED ISSUE #978: Oracle relay: pin follow-up tasks to the originating worker (session→worker affinity) so multi-account shared pools support correct multi-turn
## Summary

Add **session → worker affinity** to the Oracle relay so that a pool serving
**multiple ChatGPT accounts** can still handle multi-turn correctly. Today a
follow-up task can be dispatched to a worker on a *different* account than the one
that owns the conversation, which makes multi-turn unusable on any shared /
multi-account pool.

## Current behavior

`oracle_task_service::claim_task` claims the oldest queued task with a filter that
only matches the pool and status:

```rust
// backend/src/services/oracle_task_service.rs (claim_task)
find_one_and_update(
    doc! { "pool_id": &pool.id, "status": "queued" },
    doc! { "$set": { "status": "dispatched", "assigned_worker_id": worker_label, ... } },
    sort: { "created_at": 1 },
)
```

There is no notion of which account/worker a conversation belongs to. A follow-up
task carries a `conversation_id`; `worker_payload` resolves it to the session's
`chatgpt_url` and the worker does `goto(conversation_url)`. But a `/c/<id>`
conversation physically lives in exactly one ChatGPT account, and the server never
recorded which account/worker created it.

## Impact

- **Single-account pool** (private, or all workers on one Chrome profile/account):
  works fine — any worker can reopen any `/c/<id>` of that one account.
- **Multi-account pool** (a shared "team" pool fed by several accounts): **fresh
  tasks are fine** (competitive load balancing), but **follow-ups / `--conversation`
  / `attach` misroute**: e.g. a follow-up for account A's conversation gets claimed
  by account B's worker, which cannot open A's `/c/<id>` → wrong/blank/failed answer.

A fixed `conversation_id` pins *which conversation*, not *which account/worker*, so
this cannot be fixed client-side (a worker that declines a not-mine follow-up just
produces an `ERROR:` result, which the relay marks `failed`).

## Proposed solution (minimal, backward compatible)

Pin follow-ups to the worker/account that created the session; keep fresh tasks
competitive.

1. **Model** — add `owner_worker_label: Option<String>` to `OracleSession`
   (`models/oracle_session.rs`).
2. **Stamp owner on first answer** — in `worker_submit_result`
   (`oracle_task_service.rs`), when the session has no owner yet, set
   `owner_worker_label` to the worker that produced the first result (the account
   that created `/c/<id>`).
3. **Pin at submit + filter at claim**:
   - In `submit_task`, when creating a follow-up whose session already has an
     owner, copy it onto the task as `required_worker_label`.
   - In `claim_task`, change the queued filter to:
     ```rust
     doc! { "pool_id": &pool.id, "status": "queued",
            "$or": [ { "required_worker_label": null },                 // fresh → anyone
                     { "required_worker_label": worker_label } ] }      // follow-up → owner only
     ```

Result: fresh tasks stay load-balanced across all accounts; a follow-up is only
ever claimed by the account that owns that conversation → correct multi-turn on a
multi-account shared pool. Workers and the CDP/userscript clients need **no
change** (`required_worker_label` rides along in the existing task; the client
already obeys `conversation_url`).

## Backward compatibility / edge cases

- Legacy sessions with no `owner_worker_label` keep today's behavior (unpinned).
- If the owning account's worker is offline, its follow-ups wait in queue (correct
  — you cannot continue account A's chat without account A). Optionally add a
  lease/age fallback later.
- Add an index supporting the `required_worker_label` filter on `oracle_tasks`.

## Alternatives considered

- **One pool per account** — works today with zero code, but gives up a single
  shared entry point / cross-account load balancing for multi-turn workloads.
- **Client-side "decline if not mine"** — not viable: declined follow-ups become
  `failed`, and lease-expiry requeue is on the order of `task_timeout_secs` (4h).

## Acceptance criteria

- In a pool with workers on ≥2 distinct accounts: a multi-turn thread
  (`--new-conversation` then `--conversation <id>`) always lands back on the
  account that created it, and recalls prior context.
- Fresh single-shot tasks remain competitively load-balanced across accounts.
- Existing single-account pools and legacy sessions are unaffected.

kaihuei0114:
Closes #978.

## Problem

`oracle_task_service::claim_task` claimed the oldest queued task filtered only by pool + status, with no notion of which account/worker a conversation belongs to. On a **multi-account shared pool**, a follow-up for account A's conversation could be claimed by account B's worker, which cannot open A's `/c/<id>` → wrong/blank/failed answer. This made multi-turn unusable on any shared pool. A fixed `conversation_id` pins *which conversation*, not *which account*, so it can't be fixed client-side.

## Solution

Pin follow-ups to the worker/account that created the session; keep fresh tasks competitive.

- **Model** — `owner_worker_label: Option<String>` on `OracleSession`; `required_worker_label: Option<String>` on `OracleTask`. Both `serde(default, skip_serializing_if = None)`, so legacy docs stay unpinned and the fields never surface in consumer/worker payloads.
- **Stamp owner on first answer** — `worker_submit_result` sets `owner_worker_label` on the first *successful* result via a `{ owner_worker_label: null }` filter (idempotent → the first account to answer keeps ownership). Same stamp added to the `attach`/transcript-import path, since that account physically owns the scraped `/c/<id>`.
- **Pin at submit** — `submit_task` copies the session's owner onto follow-ups as `required_worker_label`.
- **Filter at claim** — `claim_task` queued filter now ANDs `$or: [ {required_worker_label: null}, {required_worker_label: <worker>} ]`: fresh (null/absent) → any worker; pinned follow-up → owning worker only.
- **Index** — added `{ pool_id, status, required_worker_label, created_at }` on `oracle_tasks`.

Workers and the CDP/userscript clients need **no change** — `required_worker_label` rides along in the existing task and the client already obeys `conversation_url`.

## Backward compatibility

- Legacy sessions with no `owner_worker_label` stay unpinned (today's behavior).
- If the owning account's worker is offline, its follow-ups wait in queue (correct — you cannot continue account A's chat without account A).

## Tests

Added `followup_pins_to_owning_worker`: owner stamping on first result, follow-up pinning, a non-owner worker idling rather than misrouting, and fresh tasks still claimable by any worker. Updated affected struct-literal fixtures.

> ⚠️ Not compiled locally — this environment has no Rust toolchain. Relying on CI for `cargo build`/`cargo test`.

### DIFF
--- a/backend/src/db.rs
+++ b/backend/src/db.rs
@@ -1643,6 +1643,21 @@ pub async fn ensure_indexes(db: &Database) -> Result<(), mongodb::error::Error> 
                 .build(),
         )
         .await?;
+    // Session→worker affinity: the FIFO claim filters queued tasks by
+    // `required_worker_label` (null for fresh, the owning account for
+    // follow-ups) before sorting by `created_at`.
+    oracle_tasks
+        .create_index(
+            IndexModel::builder()
+                .keys(doc! {
+                    "pool_id": 1,
+                    "status": 1,
+                    "required_worker_label": 1,
+                    "created_at": 1,
+                })
+                .build(),
+        )
+        .await?;
     oracle_tasks
         .create_index(
             IndexModel::builder()
--- a/backend/src/handlers/oracle_tasks.rs
+++ b/backend/src/handlers/oracle_tasks.rs
@@ -501,6 +501,7 @@ mod tests {
             pdf_name: Some("a.pdf".to_string()),
             conversation_id: Some("conv_1".to_string()),
             is_followup: false,
+            required_worker_label: None,
             client_ref: None,
             status: OracleTaskStatus::Queued,
             phase: None,
--- a/backend/src/models/oracle_session.rs
+++ b/backend/src/models/oracle_session.rs
@@ -37,6 +37,15 @@ pub struct OracleSession {
     /// `required_worker_label` so multi-turn lands back on the owning
     /// account in a multi-account pool. `None` for legacy/unstamped
     /// sessions (unpinned, today's behavior).
+    ///
+    /// Affinity keys on the worker *label*, so correctness rests on the
+    /// operational invariant that one stable label maps to one ChatGPT
+    /// account. Two tabs of the same account under different labels are
+    /// treated as different accounts (over-pinning — harmless beyond lost
+    /// load-balancing); two different accounts sharing a label would
+    /// reintroduce the misroute this pinning prevents. Worker label
+    /// assignment lives in the CDP/userscript clients, which already mint a
+    /// stable per-tab label (`?nyx=N` → `tab_N`).
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub owner_worker_label: Option<String>,
     pub turn_count: u64,
--- a/backend/src/models/oracle_session.rs
+++ b/backend/src/models/oracle_session.rs
@@ -32,6 +32,13 @@ pub struct OracleSession {
     /// Browser-side conversation URL pinned by the worker after turn 1.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub chatgpt_url: Option<String>,
+    /// Worker label of the account that created this conversation, stamped
+    /// on the first result. Follow-ups copy it onto their task as
+    /// `required_worker_label` so multi-turn lands back on the owning
+    /// account in a multi-account pool. `None` for legacy/unstamped
+    /// sessions (unpinned, today's behavior).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub owner_worker_label: Option<String>,
     pub turn_count: u64,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub last_task_id: Option<String>,
@@ -62,6 +69,7 @@ mod tests {
             api_key_id: Some(uuid::Uuid::new_v4().to_string()),
             tag: Some("bedc-deep".to_string()),
             chatgpt_url: Some("https://chatgpt.com/c/abc".to_string()),
+            owner_worker_label: Some("tab_1".to_string()),
             turn_count: 3,
             last_task_id: Some("task-3".to_string()),
             closed_at: None,
@@ -90,6 +98,7 @@ mod tests {
             api_key_id: None,
             tag: None,
             chatgpt_url: None,
+            owner_worker_label: None,
             turn_count: 0,
             last_task_id: None,
             closed_at: Some(Utc::now()),
@@ -111,6 +120,7 @@ mod tests {
             api_key_id: None,
             tag: None,
             chatgpt_url: None,
+            owner_worker_label: None,
             turn_count: 0,
             last_task_id: None,
             closed_at: None,
--- a/backend/src/models/oracle_task.rs
+++ b/backend/src/models/oracle_task.rs
@@ -84,6 +84,12 @@ pub struct OracleTask {
     /// True when this task continues an existing conversation.
     #[serde(default)]
     pub is_followup: bool,
+    /// When set, only the worker with this label may claim the task.
+    /// Copied from the owning session for follow-ups so multi-turn lands
+    /// back on the account that created the conversation in a
+    /// multi-account pool. `None` = fresh task, claimable by any worker.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub required_worker_label: Option<String>,
     /// Optional submitter-scoped idempotency key.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub client_ref: Option<String>,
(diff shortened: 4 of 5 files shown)


chrono-kw:
## Review: solid, well-scoped fix — okay to merge, with two things worth surfacing first

Pulled the branch and read the full `claim_task` / `submit_task` / `worker_submit_result` / `worker_payload` context, the index change, the TTL/quota model, and cross-checked every struct-literal site. The core logic is correct and genuinely backward-compatible: follow-ups pin to the account that created the conversation, fresh tasks stay competitively load-balanced.

### What's right

- **MongoDB semantics are correct.** `{ required_worker_label: null }` matches both absent and explicit-null, so fresh tasks (field omitted via `skip_serializing_if`) and legacy docs all fall into the "claimable by anyone" branch. Same trick on the `owner_worker_label: null` stamp filter gives idempotent first-writer-wins ownership. The subtle part is done right.
- **No head-of-line blocking.** The filter is applied *before* the `created_at` sort, so a follow-up pinned to an offline worker doesn't stall other workers — they skip it and claim the next matching task.
- **All 11 struct-literal sites updated** (5× `OracleSession`, 6× `OracleTask`), checked each against the diff. Should compile despite the "not compiled locally" caveat; CI's `cargo build`/`cargo test` remains the real gate.
- **Index** `{pool_id, status, required_worker_label, created_at}` supports both `$or` branches plus the sort; the old `{pool_id, status, created_at}` is retained for the other queries.
- **Test coverage is on-point**: owner stamping, follow-up pinning, non-owner idling instead of misrouting, and fresh-task load-balancing all asserted.

### Two things to flag before/with merge

**1. Pinned follow-ups whose owning worker never returns become permanent zombies that consume quota.** Chain of facts:
- `db.rs:1695` — queued tasks have no `expires_at` and are never TTL-expired by Mongo.
- A queued task is never dispatched to a non-matching worker, so its lease never starts → `requeue_expired_leases` never touches it.
- `enforce_submit_quotas` counts `status ∈ {queued, dispatched}` against **both** `per_user_max_inflight` (per submitter) and `max_queue_length` (whole pool).

So if account A's worker disappears or reconnects under a different label, A's follow-ups sit queued forever, silently eating the submitter's inflight budget and pool queue capacity with no automatic recovery. The issue explicitly defers "lease/age fallback later," so it's known — suggest tracking it as a follow-up (an age-based unpin that clears `required_worker_label`, or excluding unclaimable-pinned tasks from quota). The common single-account pool case never triggers this.

**2. Affinity keys on a client-chosen `worker_label`, not a durable account identity.** Correctness rests on the operational invariant "one stable label ⇔ one account":
- Two tabs of the *same* account under different labels (`?nyx=1` vs `?nyx=2`) are treated as different accounts → over-pinning (tab_2 could serve it but won't), and exposure to the zombie case if tab_1 closes.
- If two *different* accounts ever reuse the same label, you silently get back the exact misroute this PR fixes.

Neither is a regression, but since the fix lives or dies by that invariant, it'd be worth a line in the code/docs.

### Minor (non-blocking)

- On a **failed** first result, `turn_count` is incremented and `chatgpt_url` may be pinned (`worker_submit_result`), but `owner_worker_label` intentionally is not (correct). Narrow consequence: a failed turn-1 that still reported a URL leaves a session with a pinned URL but no owner, so the next follow-up is unpinned and may land elsewhere. Acceptable, but untested — worth a test if you want completeness.
- Owner stamping is a second `update_one` after the turn-count bump. Fine as-is (the null-guard can't be folded into the unconditional bump), just an extra round-trip.

kaihuei0114:
Thanks for the thorough read — addressed all three in `bc3353b`.

**1. Zombie pinned follow-ups consuming quota** — implemented the issue's deferred lease/age fallback as `release_stale_affinity()`: a follow-up pinned to an owning worker that never returns is unpinned once it has waited a full `task_timeout_secs` window (generous enough to absorb a tab reload / network blip), after which any worker may claim it. A non-owner that picks it up can't reopen the other account's `/c/<id>` and reports an extraction failure — but that's a terminal state that frees the inflight + queue quota instead of leaking it forever. It's swept in `claim_task`, which every live worker polls continuously, so recovery runs as long as any worker in the pool is alive. (A fully dead pool already hangs all queued tasks regardless of this PR, so no new behavior there.) Covered by `stale_followup_affinity_is_released_after_grace`.

**2. Affinity keys on a client-chosen `worker_label`** — documented the one-stable-label-⇔-one-account invariant on `OracleSession::owner_worker_label`, including the same-account-two-labels (over-pin, harmless) and two-accounts-one-label (misroute) failure modes, and noted that label assignment lives in the CDP/userscript clients (`?nyx=N` → `tab_N`).

**3. Minor — failed turn-1 with a URL but no owner** — added `failed_first_turn_leaves_session_unowned`: confirms a failed first turn counts the turn and pins `chatgpt_url` but does **not** stamp `owner_worker_label`, so the next follow-up stays unpinned.

The `{pool_id, status, required_worker_label, created_at}` index also serves the new release sweep's filter. Verified locally against MongoDB: 27/27 oracle_task tests pass, `cargo build` + `clippy` clean.

PR #1010, opened 2026-06-24
feat(oracle): deliver browser-generated images + wedge guard

kaihuei0114:
## Problem

A ChatGPT image-generation turn **wedged** an oracle task. The CDP worker's `waitForResponse` only terminated on stable assistant *text*, and an image-only turn produces none — so the task spun to `MAX_WAIT_MS` (2h), tied up one of the worker's inflight slots, and returned **neither the image nor any text**. (Reproduced: an apple-image prompt sat in `waiting_response` ~3.5 min until cancelled; the live worker log shows the same.)

## Changes

**Worker** (`integrations/oracle/cdp-worker/worker.mjs`)
- `extractImages()` pulls the latest turn's content image(s). Image-gen turns render `` inside a `conversation-turn` that carries **no** `data-message-author-role="assistant"`, so extraction scopes to the last conversation-turn (not assistant-role) and matches only ChatGPT content-host srcs (`oaiusercontent` / `backend-api` / `blob:`). Dedupes by file id so multi-resolution copies collapse.
- `waitForResponse` now returns `{ text, images }`; images carry the terminator so an image-only turn completes. Adds a **no-output idle guard** (`NO_OUTPUT_IDLE_MS`, default 420s) so a stuck/empty turn fails fast instead of wedging — mirroring the userscript.
- `downloadImages()` fetches bytes through the browser's **authenticated context** (`page.request.get` carries session cookies, bypassing the cross-origin CORS wall a page `fetch` hits), restricted to the same content-host allowlist (SSRF guard — the assistant turn is untrusted output), capped per-image and in aggregate.

**Backend**
- `OracleImage { mime, data, name }` stored as **BSON Binary** (redacting `Debug`) — image bytes live on the task doc under the same TTL/privacy discipline as response bodies, keeping the doc under the 16 MB cap.
- `/oracle/worker/result` accepts `images[]`; `worker_submit_result` validates mime + per-image/total caps. An image-only turn (empty text + ≥1 image) is treated as a **success**, not a failure. `/tasks/{id}` returns images as base64.

**CLI**
- `oracle ask` / `oracle result --out <path>` saves generated images to disk (auto-named `oracle-<task_id>-<n>.<ext>` otherwise). Image save is **non-fatal** — a failed write warns and still prints the text answer.

### PDF upload (parity note)

This branch also carries a small **PDF-attachment** path that was in-flight on the worker before this work: `handlePrompt` calls a new `uploadPdf()` when a task has `pdf_base64`, guarded by `!is_followup` so a resent attachment never re-uploads into an existing conversation (mirrors the userscript). It reuses the existing `OracleTask.pdf_base64` / `pdf_name` fields already on `main` and the same heartbeat/cancel discipline as `waitForResponse`. It is **not** the focus of this PR and is not separately covered by the E2E below — flagging it explicitly rather than letting it land silently. Happy to split it into its own PR if preferred.

## Tests / verification

- Backend oracle tests **52/0** (incl. 2 new for image validation/caps), CLI oracle tests **17/0** (incl. 3 for `--out` path resolution). `cargo fmt --all -- --check` + `cargo clippy --workspace --all-targets -- -D warnings` clean.
- **End-to-end** against a local backend + CDP worker driving a real logged-in Chrome: *"draw a red apple"* produced a **1254×1254 PNG** delivered all the way to disk via `--out`.

## Not in this PR / follow-ups

- The deployed `worker-tab-isolated.mjs` (URL-key variant) and the Ornn `nyxid-oracle-workers-setup` skill carry the same worker fix (published as skill v0.8); they live **outside this repo**.
- Backend must be **deployed** for prod image delivery; until then prod image turns fail-fast (no wedge) rather than hang.

### DIFF
--- a/backend/src/handlers/oracle_tasks.rs
+++ b/backend/src/handlers/oracle_tasks.rs
@@ -12,6 +12,7 @@ use axum::{
     http::StatusCode,
     response::IntoResponse,
 };
+use base64::Engine;
 use serde::{Deserialize, Serialize};
 
 use crate::AppState;
@@ -84,6 +85,15 @@ pub struct ExtractResponse {
 }
 
 #[derive(Serialize)]
+pub struct OracleImageInfo {
+    pub mime: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub name: Option<String>,
+    pub bytes: u64,
+    pub data_base64: String,
+}
+
+#[derive(Serialize)]
 pub struct OracleTaskInfo {
     pub task_id: String,
     pub pool_id: String,
@@ -107,6 +117,10 @@ pub struct OracleTaskInfo {
     pub response: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub response_chars: Option<u64>,
+    /// Images produced on an image-generation turn (bytes re-encoded as
+    /// base64 for JSON transport). Present only on the single-task GET.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub images: Option<Vec<OracleImageInfo>>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub chatgpt_url: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
@@ -183,6 +197,16 @@ fn task_info(task: &OracleTask, queue_position: u64) -> OracleTaskInfo {
         assigned_worker: task.assigned_worker_id.clone(),
         response: task.response.clone(),
         response_chars: task.response_chars,
+        images: task.images.as_ref().map(|imgs| {
+            imgs.iter()
+                .map(|im| OracleImageInfo {
+                    mime: im.mime.clone(),
+                    name: im.name.clone(),
+                    bytes: im.data.len() as u64,
+                    data_base64: base64::engine::general_purpose::STANDARD.encode(&im.data),
+                })
+                .collect()
+        }),
         chatgpt_url: task.chatgpt_url.clone(),
         failure_reason: task.failure_reason.clone(),
         created_at: task.created_at.to_rfc3339(),
@@ -512,6 +536,7 @@ mod tests {
             lease_expires_at: None,
             response: None,
             response_chars: None,
+            images: None,
             chatgpt_url: None,
             failure_reason: None,
             worker_script_version: None,
--- a/backend/src/handlers/oracle_worker.rs
+++ b/backend/src/handlers/oracle_worker.rs
@@ -128,10 +128,23 @@ pub async fn ack(
 }
 
 #[derive(Deserialize)]
+pub struct WorkerImageDto {
+    pub mime: String,
+    pub data_base64: String,
+    #[serde(default)]
+    pub name: Option<String>,
+}
+
+#[derive(Deserialize)]
 pub struct WorkerResultRequest {
     pub task_id: String,
     pub worker: String,
     pub response: String,
+    /// Images produced on an image-generation turn (base64). Optional and
+    /// validated/capped server-side; an image-only turn sends an empty
+    /// `response` and a non-empty `images` list.
+    #[serde(default)]
+    pub images: Option<Vec<WorkerImageDto>>,
     #[serde(default)]
     pub chatgpt_url: Option<String>,
     #[serde(default)]
@@ -152,12 +165,24 @@ pub async fn submit_result(
     Json(body): Json<WorkerResultRequest>,
 ) -> AppResult<Json<WorkerResultResponse>> {
     let pool = authenticate_worker(&state, &headers).await?;
+    let req_images = body.images.unwrap_or_default();
+    let image_count = req_images.len();
+    let image_base64_chars: usize = req_images.iter().map(|i| i.data_base64.len()).sum();
+    let images: Vec<oracle_task_service::ResultImage> = req_images
+        .into_iter()
+        .map(|i| oracle_task_service::ResultImage {
+            mime: i.mime,
+            data_base64: i.data_base64,
+            name: i.name,
+        })
+        .collect();
     let outcome = oracle_task_service::worker_submit_result(
         &state.db,
         &pool,
         &body.worker,
         &body.task_id,
         &body.response,
+        images,
         body.chatgpt_url.as_deref(),
         body.model.as_deref(),
         body.script_version.as_deref(),
@@ -165,12 +190,14 @@ pub async fn submit_result(
     )
     .await?;
 
-    // Metadata-only trace: task id + outcome + size, never the body.
+    // Metadata-only trace: task id + outcome + sizes, never the body or bytes.
     tracing::info!(
         task_id = %body.task_id,
         pool_id = %pool.id,
         outcome = ?outcome,
         response_chars = body.response.chars().count(),
+        image_count,
+        image_base64_chars,
         "Oracle worker result received"
     );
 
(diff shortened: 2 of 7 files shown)


chronoai-kai:
## Review summary

Solid core. The image-delivery path, caps layering, and privacy discipline are well done and match the oracle conventions. Two items worth addressing before merge, plus a few hardening notes.

### Please address

**1. [Bug] PDF upload is missing the `!is_followup` guard** — `worker.mjs` `handlePrompt`
The new CDP branch fires on any turn carrying a PDF:
```js
if (task.pdf_base64) { await ack(task_id, "uploading_pdf"); await uploadPdf(page, task_id, task); }
```
The reference userscript intentionally guards it (`nyxid_oracle.user.js`): `if (!is_followup && pdf_base64)`. If the server ever resends `pdf_base64` on a follow-up turn, the CDP worker will re-upload the PDF into an existing conversation while the userscript would not. Please mirror the guard: `if (!task.is_followup && task.pdf_base64)`.

**2. [Scope/coverage] PDF upload is undocumented and untested**
`main` is at `cdp-1.0` with no PDF support; this PR jumps to `cdp-1.3-image` and folds in a ~40-line `uploadPdf()` plus a `handlePrompt` branch. The PR body, commit message, and test plan are entirely about image delivery + the wedge guard and never mention PDF upload. The end-to-end verification only exercises image gen. Either split the PDF work into its own PR (cleaner history) or call it out in the description and add a parity note. As-is it lands silently.

### Hardening (non-blocking)

**3. [SSRF] `downloadImages` fetches arbitrary hosts** — `worker.mjs`
`extractImages` includes any `` with `width >= 200` even when it doesn't match the `oaiusercontent|backend-api|blob:` content pattern, then `downloadImages` does `page.request.get(src)` against it. The assistant turn is untrusted output (per "AI 产物默认不可信"), so a model-emitted `` could make the worker GET an internal/LAN URL from its network vantage and store the response (capped) as an image. Blast radius is bounded (per-domain cookie jar, size caps, needs a loadable image), but consider restricting the actual fetch to the same host allowlist `looksContent` already uses.

**4. [Robustness] image-save failure hides the text answer** — `cli/src/commands/oracle.rs`
In both `run()` arms, `save_result_images(...)?` runs before `print_result(...)`. A single undecodable image returns `Err` and aborts the command, so a text answer that arrived fine never prints. Server-side re-encoding makes this unlikely, but consider making image-save non-fatal (warn + continue) or printing the result first.

**5. [Quality] `extractImages` dedupes by exact `src`** — `worker.mjs`
ChatGPT often renders the same generated image at multiple resolutions under different URLs (thumbnail/full/zoom). Exact-src dedup won't collapse those, so the same picture can be delivered twice and burn a `MAX_IMAGES` slot. Minor.

### What's good (verified)
- Caps layered correctly: worker self-caps (4 / 6MB / 9MB) <= server re-validates (8 / 8MB / 10MB), all under the 16MB body and Mongo doc ceilings; `decode_result_images` is best-effort and tested.
- BSON Binary storage keeps the doc small; `OracleImage`'s hand-written `Debug` redacts bytes; worker trace logs only `image_count`/`image_base64_chars`; TTL rides on `expires_at`.
- No payload bloat on list paths: `task_info()` (carries base64) is only wired to `get_task`/`cancel_task`; session listing uses `OracleTurnInfo` with no images. The "single-task GET only" claim holds.
- Image-only-turn semantics are right, and the wedge guard (`NO_OUTPUT_IDLE_MS = 420s`) matches the userscript.

kaihuei0114:
Thanks — all five addressed in `4dfcf08` (worker fixes also mirrored to the deployed `worker-tab-isolated.mjs` + Ornn skill, published as **v0.8**).

**1. PDF `!is_followup` guard** — fixed. `handlePrompt` now gates on `if (!task.is_followup && task.pdf_base64)`, mirroring the userscript, so a resent attachment can't re-upload into an existing conversation.

**2. PDF undocumented/untested** — added a **"PDF upload (parity note)"** section to the PR description explaining it's a pre-existing in-flight path, reuses the `pdf_base64`/`pdf_name` fields already on `main`, and is now guarded. Flagged that it's not the focus and not separately E2E-tested here. Happy to split it into its own PR instead if you'd prefer the cleaner history — just say.

**3. SSRF in `downloadImages`** — fixed. `extractImages` now returns **only** srcs matching the content-host allowlist (`oaiusercontent` / `backend-api` / `blob:`); the `width >= 200` size-only fallback is gone, so a model-emitted `` is never extracted and therefore never fetched.

**4. image-save hides text answer** — fixed. Both `run()` arms now treat `save_result_images` as non-fatal (`warn + continue`), so a failed/undecodable image write no longer aborts before the text answer prints.

**5. exact-src dedup** — fixed. Dedup key is now the `file_<id>` token parsed from the src (falling back to exact src), so the thumbnail/full/zoom copies of one generated image collapse to a single entry instead of burning `MAX_IMAGES` slots.

Re-verified: `cargo fmt --all -- --check` + `cargo clippy --workspace --all-targets -- -D warnings` clean, CLI oracle tests 17/0, worker + DOM_CORE parse.

chronoai-kai:
## Re-review of `4dfcf08`

Thanks for the quick turnaround. Most of it is resolved:

- **PDF `!is_followup` guard** — fixed exactly, mirrors the userscript.
- **PDF scope/doc** — the new "PDF upload (parity note)" section is the right call; flagging it beats letting it land silently.
- **CLI image-save non-fatal** — fixed; a failed write now warns and the text answer still prints.
- **Dedupe by file id** — `file[-_][A-Za-z0-9]+` collapses multi-resolution copies correctly while keeping genuinely distinct images apart.

### One residual: SSRF via the `alt` branch (defense-in-depth, not a blocker)

Removing the `width >= 200` bypass closed the wide hole. Two things keep it cracked:

1. `extractImages` still authorizes via the model-controlled `alt`: `looksContent` is true on `/^generated image/i.test(img.alt)` regardless of host. A model-emitted `generated image` (e.g. assistant markdown `generated image`) passes, and a non-loading internal URL has `naturalWidth === 0` so the `< 64` size filter doesn't drop it either.
2. `downloadImages` does no host check — it `page.request.get(src)`s whatever `extractImages` returns. So the "Same allowlist downloadImages fetches" comment isn't actually enforced there, and the commit's "never an arbitrary model-emitted ``" claim doesn't fully hold.

Practical risk is low (ChatGPT proxies/sanitizes external assistant-markdown images; caps + per-domain cookie jar bound the blast radius), so this is defense-in-depth. Cleanest fix is one line at the actual trust boundary in `downloadImages`:

```js
// non-blob srcs must be a ChatGPT content host before we GET them
if (!src.startsWith("blob:") && !/oaiusercontent|backend-api/.test(src)) continue;
```

That makes the property hold regardless of the `extractImages` heuristic, and the `alt` branch can then stay or go freely.

Otherwise this looks good to me.

kaihuei0114:
Good catch on the `alt` branch — you're right that the property was being asserted in the wrong place. Fixed at the trust boundary in `d9ce7a6`, using your one-liner:

```js
// in downloadImages, before the fetch:
if (!src.startsWith("blob:") && !/oaiusercontent|backend-api/.test(src)) continue;
```

Now the "never GET an arbitrary src" guarantee holds at the point of the privileged, cookie-bearing fetch — independent of `extractImages`' heuristics. A non-loading `generated image` (naturalWidth 0, so the size filter wouldn't drop it) is matched by `extractImages` but **refused by `downloadImages`** before any request goes out. Left the `alt` branch in `extractImages` since it's now harmless and still helps find legit generated images.

Mirrored to the deployed `worker-tab-isolated.mjs` + Ornn skill (**v0.9**). `node --check` + DOM_CORE parse clean.

PR #1029, opened 2026-06-25
feat(oracle): general --attach-file upload (image / pdf / ...)

kaihuei0114:
## What

Adds a **general input-attachment** path to `oracle ask` so a vision/document model can answer questions about an uploaded file — and keeps `--pdf` fully working.

```
nyxid oracle ask <pool> "what's in this?" --attach-file diagram.png
nyxid oracle ask <pool> "summarize" --attach-file report.pdf
nyxid oracle ask <pool> "secret token?" --pdf doc.pdf      # still works
```

## Design / naming

- **`--attach-file <PATH>`**, not `--attach` — `oracle attach` is an existing **subcommand** (import a conversation by URL); reusing "attach" as a flag would be confusing. `--file` is also taken (prompt source).
- **Backward compatible:** `--pdf` and the API's `pdf_base64`/`pdf_name` are untouched — existing callers and scripts keep working. The general path uses new `attachment_base64`/`attachment_name`; the worker keeps `uploadPdf()` for the legacy field alongside the new `uploadAttachment()`.
- Mime is inferred from the filename extension (pdf, png, jpg, webp, gif, txt, csv, md, json, …; `application/octet-stream` fallback). So `--attach-file foo.pdf` also uploads a PDF.

## Changes

- **Backend** — `attachment_base64`/`attachment_name` on the task model, `SubmitOracleTaskRequest`, `SubmitTaskInput` (validated: size cap reuses the PDF byte cap; name length cap), and the `WorkerTaskPayload` dispatch DTO. `has_attachment` added to the submit audit (metadata only).
- **CLI** — `oracle ask --attach-file <PATH>`.
- **Worker** — `uploadAttachment()` (mime from extension) + first-turn-only guard; `uploadPdf()` retained for `pdf_base64`.

## Tests

- Local CI replica green: clippy `-D warnings`, backend tests, CLI tests (1032 + wizard-bundle freshness), Rust features matrix (aws-kms / gcp-kms / both). (The frontend/SDK jobs don't apply — no frontend/SDK changes.)
- **End-to-end** (local backend + CDP worker driving ChatGPT-5.5-Pro):
  - `--pdf secret.pdf` → `BANANA-7732` (back-compat, legacy `uploadPdf`)
  - `--attach-file secret.pdf` → `BANANA-7732` (PDF via general path, `application/pdf`)
  - `--attach-file diagram.png` → transcribed a 5-box / 7-arrow architecture diagram exactly

## Follow-ups (out of scope)

- The deployed `worker-tab-isolated.mjs` + Ornn skill carry the same `uploadAttachment` (for when the backend ships); skill republish is separate.
- Prod backend must be deployed for attachment input to work against prod (this branch's backend adds the field).
- Independent of PR #1028 (worker HTTPS guard) — different code regions, no conflict.

### DIFF
--- a/backend/src/handlers/oracle_tasks.rs
+++ b/backend/src/handlers/oracle_tasks.rs
@@ -41,6 +41,13 @@ pub struct SubmitOracleTaskRequest {
     pub pdf_base64: Option<String>,
     #[serde(default)]
     pub pdf_name: Option<String>,
+    /// Optional input file attachment (image / PDF / ..., base64) + filename,
+    /// uploaded by the worker on the first turn so the model can answer
+    /// questions about it. Mime is inferred from the filename extension.
+    #[serde(default)]
+    pub attachment_base64: Option<String>,
+    #[serde(default)]
+    pub attachment_name: Option<String>,
     /// Optional submitter-scoped idempotency key: retried submits with
     /// the same value return the original task.
     #[serde(default)]
@@ -259,6 +266,8 @@ pub async fn submit_task(
             conversation_id: body.conversation_id,
             pdf_base64: body.pdf_base64,
             pdf_name: body.pdf_name,
+            attachment_base64: body.attachment_base64,
+            attachment_name: body.attachment_name,
             client_ref: body.client_ref,
         },
     )
@@ -277,6 +286,7 @@ pub async fn submit_task(
             "is_followup": outcome.task.is_followup,
             "prompt_chars": outcome.task.prompt.chars().count(),
             "has_pdf": outcome.task.pdf_base64.is_some(),
+            "has_attachment": outcome.task.attachment_base64.is_some(),
             "deduplicated": outcome.deduplicated,
         })),
     );
@@ -523,6 +533,8 @@ mod tests {
             tag: None,
             pdf_base64: Some("cGRm".to_string()),
             pdf_name: Some("a.pdf".to_string()),
+            attachment_base64: None,
+            attachment_name: None,
             conversation_id: Some("conv_1".to_string()),
             is_followup: false,
             required_worker_label: None,
--- a/backend/src/handlers/oracle_worker.rs
+++ b/backend/src/handlers/oracle_worker.rs
@@ -329,6 +329,8 @@ mod tests {
                 tag: None,
                 pdf_base64: None,
                 pdf_name: None,
+                attachment_base64: None,
+                attachment_name: None,
                 required_project_url: None,
                 assigned_worker: "tab_1".to_string(),
                 submitted_at: "2026-06-11T00:00:00Z".to_string(),
--- a/backend/src/models/oracle_task.rs
+++ b/backend/src/models/oracle_task.rs
@@ -106,6 +106,15 @@ pub struct OracleTask {
     pub pdf_base64: Option<String>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub pdf_name: Option<String>,
+    /// Optional input file attachment (image / PDF / ..., base64), uploaded by
+    /// the worker on the first turn so the model can answer questions about it.
+    /// Mime is derived from `attachment_name`'s extension by the worker.
+    /// Size-capped at submit. Separate from the legacy `pdf_base64` field, which
+    /// is kept for backward compatibility.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub attachment_base64: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub attachment_name: Option<String>,
     /// Multi-turn session this task belongs to (None = single-shot).
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub conversation_id: Option<String>,
@@ -189,6 +198,8 @@ mod tests {
             tag: Some("bedc-deep".to_string()),
             pdf_base64: None,
             pdf_name: None,
+            attachment_base64: None,
+            attachment_name: None,
             conversation_id: None,
             is_followup: false,
             required_worker_label: None,
--- a/backend/src/services/oracle_task_service.rs
+++ b/backend/src/services/oracle_task_service.rs
@@ -135,8 +135,8 @@ fn validate_submit_input(input: &SubmitTaskInput) -> AppResult<()> {
             "pdf_name exceeds {MAX_PDF_NAME_LEN} chars"
         )));
     }
-    if let Some(image) = &input.attachment_base64 {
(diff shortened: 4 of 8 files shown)


chronoai-kai (commented):
Reviewed the full diff. This is clean, additive, and faithfully mirrors the existing `--pdf` path through the whole chain (model -> submit request -> `SubmitTaskInput` validation -> `WorkerTaskPayload` -> worker). Backward compatibility looks intact. A few things to push back on before this lands. None of them block merge, but #1 and #2 are real asks.

**What checks out**
- Validation is a genuine mirror and is wired in: the new branches in `validate_submit_input` (size cap, name-required-when-set, name-length cap) are reached via the call at `oracle_task_service.rs:188`, same as pdf. Size reuses `MAX_PDF_BASE64_BYTES` (12 MB) under the 16 MiB body cap.
- Audit logs only `has_attachment` (boolean), no payload, matching the oracle metadata-only discipline. Retention is covered by the existing task-doc TTL.
- Model field uses `skip_serializing_if = "Option::is_none"` (not the banned unconditional `skip_serializing`), identical to `pdf_base64`. `WorkerTaskPayload` fields mirror `pdf_name` (serialize-only, no `serde(default)` needed).
- No injection: the filename reaches `page.evaluate((fname) => ...)` as a serialized arg, not concatenated into code; `fileMime` is a fixed allowlist with an `application/octet-stream` fallback.

**1. Test coverage gap (please add).** `validate_submit_input` gained three new branches but zero new tests. The parallel pdf branches have `oversized_pdf` and `pdf_without_name` tests in the same file; the attachment branches have none (the diff only adds `attachment_base64: None` to existing fixtures). It's a literal copy so risk is low, but the repo's pattern is to test these. Please add `oversized_attachment` and `attachment_without_name` cases next to the pdf ones.

**2. Heterogeneous worker pool silently drops attachments.** Only `cdp-worker/worker.mjs` learned `uploadAttachment`. The Tampermonkey userscript (`nyxid_oracle.user.js`) and the deployed `worker-tab-isolated.mjs` do not. A task submitted with `--attach-file` that gets claimed by an un-updated worker uploads nothing and the model answers as if no file was given, which is a silently-wrong answer rather than an error. There's no backend capability gate preventing an attachment task from routing to an incapable worker. The PR body acknowledges this as a follow-up, so this is mostly a flag: make sure the prod worker rollout tracks with the backend deploy, and consider whether a capability gate is worth it before attach-file is broadly used.

**3. Naming nit.** In `validate_submit_input`, `if let Some(image) = &input.attachment_base64` binds a general attachment as `image`, but it handles pdf/txt/csv/json too. Same wording in the model doc comment ("image / PDF"). Suggest renaming the binding to `attachment` for clarity. Cosmetic.

**4. `--pdf` + `--attach-file` together is undefined.** Both flags can be set at once and the worker would upload both on the first turn. Not harmful, just unusual. Optional: clap `conflicts_with`, or leave it.

Verdict: none of these block merge. With #1 added and #2 noted in the rollout, this is good to land.

kaihuei0114:
Thanks for the thorough review. Addressed in `2edc414`:

**1. Test coverage (added).** Added `oversized_attachment` and `attachment_without_name` to `submit_validation`, right next to the pdf cases — they exercise the new size-cap and name-required branches in `validate_submit_input`. Green locally.

**2. Worker parity (fixed for the in-repo workers).** Generalized the userscript's `uploadPDF` → `uploadFile` (mime inferred from the filename) and made `nyxid_oracle.user.js` upload `attachment_base64` on the first turn — so the Tampermonkey worker now has parity with the CDP `worker.mjs`; a `--attach-file` task it claims no longer silently uploads nothing. (`.pdf` names still resolve to `application/pdf`, so the legacy pdf path is unchanged.)
   - On the **capability gate**: agreed it's the real fix for a *heterogeneous field* (an old deployed worker still no-ops). I've left that as a deliberate follow-up rather than expand this PR — the backend already records each worker's `script_version` on claim, so a gate that refuses to dispatch attachment tasks to workers below a min version is feasible. Happy to do it as its own PR; meanwhile the prod worker rollout will track the backend deploy.

**3. Naming nit (fixed).** Renamed the `if let Some(image) = &input.attachment_base64` binding to `attachment`.

**4. `--pdf` + `--attach-file` (fixed).** Added clap `conflicts_with = "pdf"` on `--attach-file`, so the two can't be set together.

Re-verified locally: `cargo fmt --all -- --check`, clippy `-D warnings`, backend `submit_validation` (with the new cases), CLI 1032 tests, userscript + worker `node --check` — all green.

chronoai-kai (approved):
Re-reviewed at `2edc4143`. The follow-up addresses all four points from the previous review.

1. **Test coverage** — fixed. `oversized_attachment` (asserts `OraclePayloadTooLarge` via `matches!`) and `attachment_without_name` now mirror the pdf tests. Ordering is correct: oversize hits the size branch first, name-missing hits the required branch.
2. **Worker pool silently dropping attachments** — fixed in-repo. `nyxid_oracle.user.js` refactored `uploadPDF` into a general `uploadFile` (mime from filename) and added the first-turn attachment block, so both in-repo workers handle it now. Clean rename, no dangling `uploadPDF` references. The deployed `worker-tab-isolated.mjs` is out-of-repo and tracked as a separate republish per the PR body.
3. **Naming** — fixed, `image` -> `attachment`.
4. **`--pdf` + `--attach-file`** — fixed, `conflicts_with = "pdf"` gives parse-time mutual exclusion.

One non-blocking note for later: the userscript's legacy pdf path now derives mime from the filename instead of hardcoding `application/pdf`, so a `pdf_name` with no `.pdf` extension would upload as `application/octet-stream`. The common path is fine (CLI passes real filenames, default `main.pdf`), so this is an edge case, not a regression. Also `fileMime` is now duplicated across `worker.mjs` and `nyxid_oracle.user.js` — acceptable given they're separate runtimes, just a minor drift risk.

LGTM. Ready to merge.

kaihuei0114:
Addressed both non-blocking notes in `a92dca7`:

1. **Legacy pdf mime** — `uploadFile` now takes an optional `mimeOverride`; the pdf path passes `"application/pdf"`, so a `pdf_name` without a `.pdf` extension still uploads as a PDF (restores the pre-refactor behavior). The attachment path keeps deriving from the filename.
2. **`fileMime` duplication** — true de-dup isn't feasible (the userscript can't import and the worker ships as one self-contained `.mjs`), so I added cross-reference "keep in sync" comments to both copies to flag the drift risk.

JS-only change (userscript + `worker.mjs`); Rust/CI unaffected.

PR #1099, opened 2026-07-07
fix: resolve service-account roles as first-class principals

### LINKED ISSUE #1098: Service account role_ids never resolve into RBAC (proxy identity assertion, introspection, userinfo)
## Background

NyxID lets an admin assign RBAC roles to a **service account**:

```
nyxid service-account create --name ci-bot --scopes "openid profile proxy" --role-ids <role>
# stored as service_accounts.role_ids, returned by GET /api/v1/admin/service-accounts/{id}
```

However, **no RBAC path in NyxID ever reads `service_accounts.role_ids`.** Every path that resolves a principal's roles/permissions — the proxy identity assertion, `/oauth/introspect`, `/oauth/userinfo`, and access-token claim injection — funnels through `services/rbac_helpers.rs::resolve_user_rbac(db, user_id)`, which looks the id up **only in the `users` collection**. A service account has no `users` document, so `resolve_user_rbac` returns empty and the assigned `--role-ids` are silently ignored everywhere.

The documented `--role-ids` feature is therefore a no-op for every consumer, independent of any particular downstream service.

## Impact

A downstream service behind the NyxID proxy that is configured with `identity_propagation_mode = jwt` (or `headers`/`both`) receives **no roles/permissions** for a service-account caller:

- `handlers/proxy.rs::execute_proxy_inner` builds the `X-NyxID-Identity-Token` assertion (and `X-NyxID-User-Roles/Permissions/Groups` headers) **only inside `if let Some(user) = users.find_one(id)`**. For an SA (no users doc) the whole block is skipped → **no identity token is injected at all** → the downstream 401s the SA.
- `/oauth/introspect` reports an active SA token with `roles: []`, `permissions: []` — contradicting its assigned `role_ids`.
- `/oauth/userinfo` hard-404s an SA access token (`users.find_one(...).ok_or(NotFound)`).

Concretely: an Ornn backend service account cannot upload to a `chrono-bucket`/`chrono-storage`-style downstream service, because its role never reaches the service.

### The only current workaround is a manual DB insert

The service accounts that *do* work today (`Ornn`, `Sisyphus`) each have a hand-seeded `users` document whose `_id` equals the SA id (with a synthetic `sa-*@nyxid.local` email). No API produces that: `POST /admin/users` mints its own random id and requires email+password; `assign_role` requires a pre-existing users doc; SA-create never writes a users doc. So the current "fix" requires direct Mongo access — not viable in production.

## Root cause

Two layers assume a human `users` document when resolving a **principal**:

1. `services/rbac_helpers.rs::resolve_user_rbac` — reads only `users`, returns empty on miss (so SA `role_ids` never resolve).
2. `handlers/proxy.rs` (and its MCP twin `services/mcp_service.rs`) — gate identity emission on `users.find_one(id)` succeeding, so an SA never reaches `identity_service::generate_identity_assertion` at all.

`service_account.role_ids` is stored and returned by the admin API but read by **no** RBAC code path.

## Proposed fix — "service accounts are first-class principals" (single source of truth)

Resolve a principal's roles from **whichever record owns the id** (users *or* service_accounts); keep role ids on their native record (no users-doc seeding, no data duplication, no migration). Scope:

1. `resolve_user_rbac`: on a `users` miss, fall back to `service_accounts` and resolve `sa.role_ids` through the identical role-fetch/flatten (SAs have no groups). This transparently fixes introspection and any id-keyed consumer.
2. `handlers/proxy.rs`: resolve a `Principal` (user **or** service account, branching on the verified `AuthMethod::ServiceAccount`) and emit the identity headers/assertion for it. Identity subject = the SA's own id (`auth_user.user_id`), never `proxy_resolution_user_id()` (the owner).
3. `identity_service.rs`: a small `Principal` seam (`From<&User>` / `From<&ServiceAccount>`) so the header + assertion builders work for either without assuming a `users` doc.

**Explicitly out of scope / untouched:** the `api_v1_human_only` fence (SAs stay blocked from management endpoints), and SA access-token issuance (`generate_service_account_token` keeps `roles: None` — request-time resolution is by design).

## Backward compatibility

The only behavior change is on the proxy path for services with `identity_propagation_mode != none` — the default is `none` everywhere, so the default fleet is unaffected. Services that today implicitly reject SA calls (no identity token) will start receiving a proper SA identity token; that is the intended fix, not a regression.

## Definition of Done

- SA `--role-ids` resolves into the proxy identity assertion and `/oauth/introspect`.
- No DB seeding required; human RBAC path byte-identical.
- Human-only fence intact; existing tests green.
- Verified end-to-end: an SA with an admin role can call a proxied downstream service; an SA with no role is denied.

chronoai-shining:
> **Updated after review:** the MCP transport path and `/oauth/userinfo` are now principal-aware for service accounts, unit tests added for the SA resolution + `Principal::from(&ServiceAccount)`, and rustfmt applied. All CI green. See the review reply for the point-by-point response.

---

## Summary

Service accounts can be assigned RBAC roles (`nyxid service-account create --role-ids …`, stored on `service_accounts.role_ids`), but **no RBAC path ever read them** — every path resolves roles via `resolve_user_rbac(db, user_id)`, which looks the id up only in the `users` collection, and a service account has no users doc. The result: the proxy identity assertion, `/oauth/introspect`, and `/oauth/userinfo` all returned **empty** roles/permissions for a service-account caller, so a downstream service behind the proxy never received the SA's roles and rejected it.

This makes RBAC + identity propagation **principal-aware**: a principal's roles resolve from whichever record owns the id (users *or* service_accounts). Single source of truth — no users-doc seeding, no data duplication, no migration.

Closes #1098.

## Root cause

Two layers assumed a human `users` document when resolving a **principal**:
1. `services/rbac_helpers.rs::resolve_user_rbac` — reads only `users`, returns empty on miss.
2. `handlers/proxy.rs` — builds the identity assertion/headers **only inside `if let Some(user) = users.find_one(id)`**, so an SA (no users doc) is skipped entirely and gets no `X-NyxID-Identity-Token`.

`service_account.role_ids` was stored and returned by the admin API but read by no RBAC code path.

## Implementation

**1. `services/rbac_helpers.rs` — the resolver seam.**
`resolve_user_rbac` now: looks up `users`; on a miss, falls back to `service_accounts` and resolves `sa.role_ids` (SAs have no groups). Extracted the role-fetch/flatten machinery into a shared `resolve_rbac_from_ids(role_ids, group_ids)` so the human path is byte-identical and both principal types share it. This single change also fixes `/oauth/introspect` (its `claims.sub` is the SA id) with no other edit.

**2. `services/identity_service.rs` — the `Principal` seam.**
Added `Principal { id, email: Option, display_name: Option, role_ids, group_ids }` with `From<&User>` / `From<&ServiceAccount>` (SA → `email: None`, `display_name: name`, `group_ids: []`). `build_identity_headers` and `generate_identity_assertion` now take `&Principal` (email push guarded on the `Option`) and resolve RBAC from the principal's own ids — removing the second, users-only chokepoint inside the assertion builder.

**3. `handlers/proxy.rs` — emit identity for SAs.**
The identity block now resolves a `Principal` for a **user OR a service account**, branching on the verified `AuthMethod::ServiceAccount` and keying on the SA's own id (`auth_user.user_id`, never `proxy_resolution_user_id()` which is the owner). SAs now reach `build_identity_headers` / `generate_identity_assertion` instead of being silently skipped; an `else` branch warns instead of dropping identity silently.

**4. `services/mcp_service.rs` — mechanical.**
Wraps the existing user in `Principal::from(user)`. The MCP path sits behind the `api_v1_human_only` fence (SA-unreachable), so behavior is unchanged; this is only to keep the shared builders' signature consistent.

### Out of scope (deliberately untouched)
- The `api_v1_human_only` fence + `reject_service_account_tokens` — SAs stay blocked from all management endpoints.
- SA access-token issuance (`generate_service_account_token` keeps `roles: None`) — request-time resolution is by design.

## Backward compatibility

The only behavior change is on the proxy path for services with `identity_propagation_mode != none`. The default is `none` everywhere, so the default fleet is unaffected. Services deliberately configured for identity propagation will start receiving a proper SA identity token where they previously received none — the intended fix, not a regression (nothing relies on "SA denied via a missing token").

## Testing

- `cargo check` clean on `main` (v0.7.1).
- Existing unit tests green (`identity_service` 9/9, header sanitization 6/6, rbac tests).
- **End-to-end, live**, through the real proxy into a downstream resource server (chrono-bucket) with three service accounts:
  - SA with **no** role → downstream denies every operation (403).
  - SA with a **user** role → owns/operates its own resources; admin-only ops denied.
  - SA with an **admin** role → full access incl. list + ownership bypass.
  - 26/26 security assertions passed. The decoded injected identity token carries `roles:["chrono-bucket-admin"]` for the admin SA — previously `roles` was absent entirely.

## Note

Deployed and verified from a feature branch (`v0.4.0`); this PR is rebased onto `main` (`v0.7.1`) and compiles cleanly there. The 4 changed files were untouched by intervening `main` work.

### DIFF
--- a/backend/src/handlers/oauth.rs
+++ b/backend/src/handlers/oauth.rs
@@ -12,9 +12,10 @@ use crate::AppState;
 use crate::errors::{AppError, AppResult};
 use crate::handlers::admin_helpers::{extract_ip, extract_user_agent};
 use crate::models::authorization_code::{ExternalSubjectRef, validate_external_subject_params};
+use crate::models::service_account::{COLLECTION_NAME as SERVICE_ACCOUNTS, ServiceAccount};
 use crate::models::service_account_token::{COLLECTION_NAME as SA_TOKENS, ServiceAccountToken};
 use crate::models::user::{COLLECTION_NAME as USERS, User};
-use crate::mw::auth::{AuthUser, OptionalAuthUser};
+use crate::mw::auth::{AuthMethod, AuthUser, OptionalAuthUser};
 use crate::services::{
     audit_service, consent_service, oauth_broker_service, oauth_client_service, oauth_service,
     par_service, service_account_service, social_token_exchange_service, token_exchange_service,
@@ -1664,18 +1665,14 @@ pub async fn userinfo(
     auth_user: AuthUser,
 ) -> AppResult<Json<UserinfoResponse>> {
     let user_id_str = auth_user.user_id.to_string();
-    let user = state
-        .db
-        .collection::<User>(USERS)
-        .find_one(doc! { "_id": &user_id_str })
-        .await?
-        .ok_or_else(|| AppError::NotFound("User not found".to_string()))?;
 
     // Check scopes from the access token claims
     let scopes: Vec<&str> = auth_user.scope.split_whitespace().collect();
     let include_roles = scopes.contains(&"roles");
     let include_groups = scopes.contains(&"groups");
 
+    // RBAC resolution is principal-aware (users OR service_accounts) so it works
+    // for both a human user and a service-account subject.
     let (roles, groups, permissions) = if include_roles || include_groups {
         let rbac =
             crate::services::rbac_helpers::resolve_user_rbac(&state.db, &user_id_str).await?;
@@ -1699,6 +1696,34 @@ pub async fn userinfo(
     } else {
         (None, None, None)
     };
+
+    // Service accounts have no users doc; build the subject from the SA record so
+    // an SA access token gets a 200 (with SA-resolved roles) instead of a 404.
+    if auth_user.auth_method == AuthMethod::ServiceAccount {
+        let sa = state
+            .db
+            .collection::<ServiceAccount>(SERVICE_ACCOUNTS)
+            .find_one(doc! { "_id": &user_id_str })
+            .await?
+            .ok_or_else(|| AppError::NotFound("Service account not found".to_string()))?;
+        return Ok(Json(UserinfoResponse {
+            sub: sa.id,
+            email: None,
+            email_verified: None,
+            name: Some(sa.name),
+            picture: None,
+            roles,
+            groups,
+            permissions,
+        }));
+    }
+
+    let user = state
+        .db
+        .collection::<User>(USERS)
+        .find_one(doc! { "_id": &user_id_str })
+        .await?
+        .ok_or_else(|| AppError::NotFound("User not found".to_string()))?;
 
     Ok(Json(UserinfoResponse {
         sub: user.id.to_string(),
--- a/backend/src/handlers/proxy.rs
+++ b/backend/src/handlers/proxy.rs
@@ -13,10 +13,11 @@ use utoipa::ToSchema;
 
 use crate::AppState;
 use crate::errors::{AppError, AppResult};
+use crate::models::service_account::{COLLECTION_NAME as SERVICE_ACCOUNTS, ServiceAccount};
 use crate::models::service_billing::{BillingMetric, PlatformUsage, ResaleUsage};
 use crate::models::usage_meter::CredentialClass;
 use crate::models::user::{COLLECTION_NAME as USERS, User};
-use crate::mw::auth::AuthUser;
+use crate::mw::auth::{AuthMethod, AuthUser};
 use crate::services::node_ws_manager::{NodeProxyRequest, ProxyResponseType, StreamChunk};
 use crate::services::{
     approval_service, audit_service, chatgpt_translator, delegation_service, identity_service,
@@ -1671,18 +1672,35 @@ async fn execute_proxy_inner(
     let mut identity_headers = Vec::new();
 
     if target.service.identity_propagation_mode != "none" {
-        let user = state
-            .db
-            .collection::<User>(USERS)
-            .find_one(doc! { "_id": &user_id_str })
-            .await?;
+        // Resolve the propagation principal — a human user OR a service account —
+        // so identity is emitted for both. A service account has no `users` doc;
+        // detect it via the verified auth method and load the SA record instead.
+        let principal: Option<identity_service::Principal> =
+            if auth_user.auth_method == AuthMethod::ServiceAccount {
+                state
+                    .db
+                    .collection::<ServiceAccount>(SERVICE_ACCOUNTS)
+                    .find_one(doc! { "_id": &user_id_str })
+                    .await?
+                    .as_ref()
+                    .map(identity_service::Principal::from)
+            } else {
+                state
+                    .db
+                    .collection::<User>(USERS)
+                    .find_one(doc! { "_id": &user_id_str })
+                    .await?
+                    .as_ref()
+                    .map(identity_service::Principal::from)
+            };
 
-        if let Some(ref user) = user {
+        if let Some(ref principal) = principal {
             if matches!(
                 target.service.identity_propagation_mode.as_str(),
                 "headers" | "both"
             ) {
-                identity_headers = identity_service::build_identity_headers(user, &target.service);
+                identity_headers =
+                    identity_service::build_identity_headers(principal, &target.service);
             }
 
             if matches!(
@@ -1692,7 +1710,7 @@ async fn execute_proxy_inner(
                 match identity_service::generate_identity_assertion(
                     &state.jwt_keys,
                     &state.config,
-                    user,
+                    principal,
                     &target.service,
                     &state.db,
                 )
@@ -1710,6 +1728,12 @@ async fn execute_proxy_inner(
                     }
                 }
             }
+        } else {
+            tracing::warn!(
+                service_id = %service_id,
+                user_id = %user_id_str,
+                "No principal found for identity propagation"
+            );
         }
 
         match crate::services::rbac_helpers::resolve_user_rbac(&state.db, &user_id_str).await {
--- a/backend/src/services/identity_service.rs
+++ b/backend/src/services/identity_service.rs
@@ -249,6 +249,53 @@ mod tests {
         }
     }
 
+    fn make_sa() -> crate::models::service_account::ServiceAccount {
+        crate::models::service_account::ServiceAccount {
+            id: "sa-123".to_string(),
+            name: "CI Bot".to_string(),
+            description: None,
+            client_id: "sa_client".to_string(),
+            client_secret_hash: "hash".to_string(),
+            secret_prefix: "sas_1234".to_string(),
+            role_ids: vec!["role-1".to_string()],
+            allowed_scopes: "openid profile".to_string(),
+            is_active: true,
+            rate_limit_override: None,
+            created_by: "admin-1".to_string(),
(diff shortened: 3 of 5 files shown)


chronoai-kai:
Reviewed this against #1098 by verifying every claim in the issue on `main`. The issue's diagnosis is accurate (confirmed `resolve_user_rbac` users-only miss, the `users.find_one` gate in `execute_proxy_inner`, introspect returning empty roles, userinfo 404, SA tokens being accepted on the proxy router, and `role_ids` having zero RBAC consumers), and the fix here matches the proposed design. Three gaps before this should land:

**1. The MCP twin is still broken (and now inconsistently so).**

Issue #1098's root-cause section names `services/mcp_service.rs` as having the same `users.find_one` gate, and service accounts *do* reach MCP tool execution: `authenticate_mcp` in `handlers/mcp_transport.rs` explicitly verifies `claims.sa == Some(true)` tokens against `service_accounts` and uses the SA id as `user_id`. This PR only mechanically wraps the found user in `Principal` inside `execute_tool` — the SA branch was not added there. Post-merge behavior on the MCP path for an SA against a `jwt`/`both` service:

- `X-NyxID-User-Roles` / `X-NyxID-User-Permissions` **are** now emitted (that block is outside the user gate and picks up the new `resolve_user_rbac` fallback), but
- `X-NyxID-Identity-Token` is still never generated (users-doc miss skips the whole assertion block),

so a downstream validating the assertion still 401s the SA via MCP — the exact failure this PR fixes on the proxy path — and it now receives roles headers without an identity token, which is a confusing half-identity.

Suggested fix, no signature plumbing needed: `execute_tool` doesn't know the auth method, so mirror the `resolve_user_rbac` fallback instead of branching on `AuthMethod` — on `users` miss, try `service_accounts.find_one` and build `Principal::from(&sa)`. If MCP is intentionally out of scope, please say so explicitly in the PR description and file a follow-up, because as written the PR closes an issue whose root-cause section names this path.

**2. No tests for the new behavior.**

All test changes in the diff are mechanical `Principal::from(&user)` rewrites of existing user cases. Missing at minimum:
- `resolve_user_rbac` resolves an SA's `role_ids` into slugs/permissions on a users miss;
- id in neither collection still returns empty (the existing `resolve_user_rbac_missing_user_returns_empty` covers this only implicitly — worth keeping green deliberately);
- a `Principal::from(&ServiceAccount)` header-building case (no email header emitted even when `identity_include_email = true`).

The issue's DoD also calls for the e2e check (SA with role can call a proxied downstream; SA without role is denied) — worth noting in the PR description how that was verified.

**3. `/oauth/userinfo` remains a hard 404 for SAs.**

The issue's title and impact section list userinfo, but neither the issue's fix scope nor this PR touches it (`userinfo` still does `users.find_one(...).ok_or(NotFound)` before any RBAC resolution, so the fallback never runs). Probably the right call — OIDC userinfo is an end-user endpoint — but the PR description should state it explicitly so the issue isn't closed with userinfo silently unfixed.

Also verified there's no privilege-escalation surface from broadening `resolve_user_rbac`: `/admin` is nested in the human-only router behind `reject_service_account_tokens`, so an SA with an admin role still can't reach admin endpoints; and the default `identity_propagation_mode = "none"` keeps the existing fleet's behavior unchanged, as the issue claims.

One nit on the issue text itself (doesn't affect this PR): the roles/permissions/groups header block in both `proxy.rs` and `mcp_service.rs` is *outside* the `if let Some(user)` gate, not inside it as #1098 states — the headers are absent for SAs because `resolve_user_rbac` returned empty, not because the block is skipped. Same net effect, and it's exactly why fix (1) alone lights up the roles headers on both paths.

chronoai-shining:
Thanks for the very thorough review — all three points are addressed (commit `c877900`, rustfmt in `18392c7`), and CI is green.

**1. MCP path — fixed.** You're right, and my audit missed it: the `/mcp` *transport* route (`routes.rs:923`, `mcp_post` → `execute_tool`) is **not** behind the human-only fence, and `authenticate_mcp` accepts `claims.sa` (resolving the SA id). `execute_tool` now mirrors the `resolve_user_rbac` fallback exactly as you suggested — no `AuthMethod` plumbing: on a `users` miss it loads the `ServiceAccount` and builds `Principal::from(&sa)`, so the identity token is now emitted for an SA on the MCP path too. No more roles-headers-without-identity-token half-identity.

**2. Tests — added.**
- `resolve_user_rbac` resolves an SA's `role_ids` via the users-miss fallback (`rbac_helpers`).
- an id in neither collection returns empty (explicit test; the existing missing-user test kept green deliberately).
- `Principal::from(&ServiceAccount)` shape, and `build_identity_headers` omits the email header even when `identity_include_email = true` (`identity_service`).

The DoD e2e was verified live through the proxy into a downstream resource server: an SA with `chrono-bucket-admin` passes the downstream's permission + ownership checks (decoded assertion carries `roles:["chrono-bucket-admin"]`); an SA with no role is denied (403); and a **human** user with the role gets the same access, revocable on role removal (403 → 200 → 403).

**3. `/oauth/userinfo` — fixed** (rather than documented out-of-scope, since the issue names it). It's now principal-aware: an SA access token returns `200` with `sub = SA id`, `name = SA name`, no email, and SA-resolved roles/permissions — consistent with introspection. Happy to scope it back to end-user-only if you'd prefer.

Also confirmed your nit on the *issue* text: the roles/permissions/groups header block is outside the `if let Some(user)` gate — same net effect, and it's exactly why the resolver fallback alone lights up the roles headers on both the proxy and MCP paths.

chronoai-kai:
Re-reviewed commit `c877900` (+ rustfmt `18392c7`) — all three points are properly addressed. Verified against the actual diff, not just the description:

1. **MCP path**: `execute_tool` now does the users-miss → `service_accounts` fallback and builds `Principal::from(&sa)`, so the identity assertion is emitted for SAs on MCP too. Confirmed it mirrors the proxy branch behaviorally (users-first precedence makes the missing `AuthMethod` check equivalent). The roles-headers-without-token half-identity is gone.
2. **Tests**: `resolve_user_rbac_resolves_service_account_roles`, `resolve_user_rbac_unknown_principal_returns_empty` (fail-closed), `service_account_principal_shape`, and `service_account_headers_omit_email_even_when_requested` cover exactly what was missing. Backend CI + coverage gate green.
3. **userinfo**: the SA branch is correct and reachable — `/oauth` is nested in the top-level router (routes.rs:1086), outside the `/api/v1` `reject_service_account_tokens` layer, and the `AuthUser` extractor resolves `sa=true` tokens. Returning `sub`/`name`/roles for the SA's own valid token adds no new exposure, and keeping it consistent with introspection is the better call than scoping it out — no objection.

Two non-blocking nits, fine to leave as-is or fold in if you touch this again:
- proxy.rs logs a `warn` when no principal is found for propagation; the MCP twin silently skips. Trivial inconsistency.
- The userinfo SA branch has no handler-level test (covered by your live e2e; coverage delta is 0.00 so no gate pressure).

LGTM.

PR #1151, opened 2026-07-15
Support in-place broker grant review

eanz17:
## Problem

Aevatar Lark bindings need to review and expand an existing NyxID service grant without replacing the opaque binding credential. Historical bindings may contain only the `aevatar` service while runtime calls also require the configured LLM and Ornn services.

Tracks https://github.com/aevatarAI/aevatar/issues/2754.

## Solution

- Carry `binding_grant_id=SHA-256(binding_id)` through authorize, PAR, signed consent, authorization code, and token exchange.
- Revalidate the authenticated user, OAuth client, and exact external subject against the active binding.
- Display current, required, new, and optional service access in the existing OAuth consent picker.
- Keep required services selected while allowing optional grants to be added or removed.
- Optimistically rotate the refresh grant behind the existing binding and return `binding_updated=true` without issuing a replacement `binding_id`.
- Document the browser-safe in-place review protocol.

## Verification

- `cargo check -p nyxid`
- broker grant tests: 4 passed
- OAuth consent tests: 26 passed
- frontend TypeScript check passed
- `git diff --check`

Browser screenshot checks remain pending because the in-app browser runtime fails to initialize with `Cannot redefine property: process`.

### DIFF
--- a/backend/src/errors/mod.rs
+++ b/backend/src/errors/mod.rs
@@ -88,6 +88,12 @@ pub enum AppError {
     #[error("Invalid target: {0}")]
     InvalidTarget(String),
 
+    #[error("Required service is not connected: {service_name} ({service_slug})")]
+    RequiredServiceNotConnected {
+        service_slug: String,
+        service_name: String,
+    },
+
     #[error("Role not found: {0}")]
     RoleNotFound(String),
 
@@ -411,7 +417,8 @@ impl AppError {
             | Self::InvalidDpopProof(_)
             | Self::InvalidRedirectUri
             | Self::InvalidScope(_)
-            | Self::InvalidTarget(_) => StatusCode::BAD_REQUEST,
+            | Self::InvalidTarget(_)
+            | Self::RequiredServiceNotConnected { .. } => StatusCode::BAD_REQUEST,
             Self::RoleNotFound(_) | Self::GroupNotFound(_) | Self::ConsentNotFound => {
                 StatusCode::NOT_FOUND
             }
@@ -542,6 +549,7 @@ impl AppError {
             Self::InvalidScope(_) => 3002,
             Self::InvalidTarget(_) => 3005,
             Self::InvalidDpopProof(_) => 3006,
+            Self::RequiredServiceNotConnected { .. } => 3007,
             Self::RoleNotFound(_) => 4000,
             Self::GroupNotFound(_) => 4001,
             Self::ConsentNotFound => 4002,
@@ -652,7 +660,7 @@ impl AppError {
             Self::PkceVerificationFailed | Self::InvalidRedirectUri => "invalid_grant",
             Self::InvalidDpopProof(_) => "invalid_dpop_proof",
             Self::InvalidScope(_) => "invalid_scope",
-            Self::InvalidTarget(_) => "invalid_target",
+            Self::InvalidTarget(_) | Self::RequiredServiceNotConnected { .. } => "invalid_target",
             Self::Unauthorized(_)
             | Self::AuthenticationFailed(_)
             | Self::ServiceAccountNotFound(_)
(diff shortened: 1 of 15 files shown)


eanz17:
Follow-up `2db4a99a` replaces the consent page nested `URLSearchParams` memos with one immutable mount-time query snapshot.

Verification:

- `npx eslint src/pages/oauth-consent.tsx`: passed
- OAuth consent tests: 26 passed
- `npm run build`: passed
- `git diff --check`: passed

The first CI run reported full-repository lint errors. All remaining error paths are unchanged from `origin/main` (`auth-flows.tsx`, `admin-invite-codes.tsx`, `approval-grants.tsx`, `keys.tsx`, `nodes.tsx`, `org-detail.tsx`, and `service-detail.tsx`); `git diff origin/main...HEAD` contains only `oauth-consent.tsx` among the reported paths, and that changed file now lints cleanly. Those unrelated baseline errors are intentionally not folded into this grant-review PR.

eanz17:
Follow-up `01797ae7` completes the missing-required-service path:

- known, active, user-visible catalog services that are not connected now produce a typed `RequiredServiceNotConnected` domain error
- OAuth token wire responses remain RFC-compatible `invalid_target`
- browser authorization redirects carry a dedicated `required_service_not_connected` key plus encoded service identity
- the error page links to the existing `/keys?tab=services&slug=...` connection flow and offers authorization retry
- malformed, unknown, inactive, or invisible resources still use the generic `invalid_target` path
- a backend protocol test covers authorize -> consent -> code exchange -> binding token exchange for `aevatar`, `chrono-llm-public`, and `ornn-api`

Verification for the new head is complete. PR CI is fully green, including Backend Test, backend/frontend coverage, Rust format/clippy, all KMS feature builds, Frontend, and CLI wizard bundle freshness. Local locked-dependency verification also passed: frontend lint with 0 errors, 152 test files / 1,710 tests, production build, `cargo check -p nyxid`, and Rust format.

Browser screenshots are still blocked by the in-app Browser runtime import error `Cannot redefine property: process`; no external Playwright fallback was used.

chronoai-kai:
Reviewed the full diff (backend broker/oauth/errors/models/resource + frontend consent/error). Design is solid: the SHA-256 hash (not the raw `binding_id`) is what travels through the browser, `resolve_binding_grant` re-validates user + client + exact external subject + active refresh on every resolve, the top-level `broker_require_sender_constraint` gate covers the update path too, and `update_binding_grant` uses a proper optimistic CAS on `refresh_token_jti` + `rotation_version` with a concurrent-update test. Enum/error plumbing for `RequiredServiceNotConnected` (3007) is complete across status/error_code/oauth_error/error_key and the frontend labels. No correctness blockers.

One thing to confirm before merge (product behavior, not a bug):

**RFC 8707 `resource` params are now mandatory in the general consent flow, not just broker review.** Resource-derived services render as disabled/required checkboxes and are force-added server-side in `authorize_decision`, and requesting a resource for a service the user hasn't connected now hard-fails with `RequiredServiceNotConnected` + the "Connect service" screen. The previous behavior where a user could deselect a requested resource was removed (the `submits only the selected requested resources` test is gone). This is coherent, but it changes every resource-scoped authorize, not only the new binding-review path. Please confirm removing partial-decline of requested resources is intended.

Two optional nits (non-blocking):

1. `services/oauth_broker_service.rs` (retry-exhaustion cleanup, ~line 191): revokes `refresh_token_jti` unconditionally. In the non-`offline_access` case `binding_refresh` reuses the exchange's refresh token, so a 3x-conflict (rare) would revoke that token. Harmless in practice since it isn't attached to the binding and the request errors anyway, just slightly surprising. A short comment noting the intent would help.
2. `handlers/oauth.rs` (`authorize_decision`, ~line 148): the `resolve_binding_grant_review` result is re-resolved for validation then discarded. Good defensive re-check; the only cosmetic effect is the "Authorized now" badges could be stale if the grant changed between authorize and decision. No correctness impact.

Tests and CI look great: full authorize->consent->code->binding-exchange e2e, concurrent rotation, ownership-mismatch, in-place rotation, plus frontend binding-review and error-page coverage.

PR #1187, opened 2026-07-17
fix(assistant): drop racing client-side guard so /assistant stops flickering

ctkm-aelf:
## Problem

The `/assistant*` routes were gated twice, by two guards reading different data at different times:

1. **`assistantBeforeLoad`** (router) — server-verified via a fresh `/users/me`, blocks the route transition until the decision is made. Correct.
2. **`AssistantRouteGuard`** (component) — re-evaluated the auth-store snapshot on every render, returned `null` whenever `isLoading` was true, and redirected to `/dashboard` the instant it saw `isLoading === false` with the flag missing.

Guard 2 raced every transient store state, which is the flicker/bounce users see:

- On a hard load, `assistantBeforeLoad`'s own `/users/me` fetch admits the user (`setUser` — which never clears `isLoading`), but the boot `checkAuth()` can still be in flight → the already-admitted page renders `null` (blank flash) until the second fetch settles.
- `apiClient` calls `setUser(null)` on any 401 → the store momentarily reads `{user: null, isLoading: false}` → instant client-side bounce to `/dashboard` even though the server-side check had just passed — i.e. "the route guard kicks in before the permission is loaded".
- Any mid-session `checkAuth()` (onboarding refresh does this) flips `isLoading` to true → the entire assistant surface unmounts, then remounts.

Reproduced the blank first paint on the base commit via headless-browser QA before fixing.

## Fix

- **Delete `AssistantRouteGuard`** (FI-007). The navigation-blocking, server-verified `assistantBeforeLoad` is the single authority; the previous page stays on screen while it resolves, so there is no intermediate frame to flicker and no snapshot race. Flag revocation mid-session still locks the route on the next navigation (the documented contract on `fetchAssistantAccessUser`), and the backend authorizes every API call regardless.
- **Skip re-verification on `cause === "stay"` re-runs** so switching conversations (`?c=` search changes) no longer awaits a `/users/me` round trip per click. Route entries (`enter`/`preload`) still verify against the server.
- **Fall back to the authenticated store snapshot on a transient (non-401) verification failure** instead of bouncing an entitled user to `/dashboard`; still fail-closed when neither source shows the flag. 401s keep clearing the session and routing to login.

`shouldRedirectFromAssistant` is replaced by `hasAssistantAccess(fetchedUser, snapshotUser)` which encodes the fetched-wins/fallback rule and is unit-tested.

## Verification

- `npm run build` (tsc -b), `npm run lint`, full vitest suite: 169 files / 1824 tests pass.
- Headless browser QA in mock mode: direct hard-load of `/assistant` renders the full shell immediately (no blank frame, no bounce, no console errors); conversation switching and Plugins/Approvals navigation work.
- Wizard bundle regenerated — byte-identical, so no `cli/src/wizard` changes.

## Note (pre-existing, not addressed here)

`/assistant/approvals` in **mock mode** crashes `ApprovalsView` with `Cannot read properties of undefined (reading 'localeCompare')` (`approvals-view.tsx:216` — mock `/approvals` rows lack `expires_at`/timestamps). Confirmed present on the base commit before this change; worth a separate small fix.


ctkm-aelf:
**Independent review (Codex / GPT-5.6-Sol) + hardening follow-up (6cc1e082)**

An independent codex review of the original fix confirmed the reported flicker/bounce is resolved and raised 1 critical + 3 advisory findings. Disposition:

- **[P1] fixed** — a late `/users/me` 200 could resurrect a session that a concurrent 401 had cleared mid-await (this ordering pre-dated the original fix). The beforeLoad decision now lives in `resolveAssistantEntry()`: the fetched user is applied only if the session is still authenticated and the user id matches; otherwise the response is discarded and entry redirects to login (sign-out wins).
- **[P2] fixed** — `cause === "stay"` re-runs now do a network-free check of the settled store snapshot, so a mid-session flag revocation (or dead session) evicts on the next conversation switch instead of lingering. `isLoading` never triggers a redirect, so the original race cannot re-enter through this path.
- **[P2] fixed** — the lifecycle now has direct unit tests (+13): enter/preload/stay, 401-mid-fetch ordering, user-switch mid-fetch, snapshot fallback, fail-closed defaults.
- **[P2] accepted** — with `defaultPreload: "intent"`, a hover-time preload decision can be reused at click within the preload staleness window (seconds-wide, UI-freshness only; the backend independently authorizes every assistant API call).

Local gates after the follow-up: build ✓, lint ✓, 1837 tests ✓, wizard bundle byte-identical.

PR #1199, opened 2026-07-17
fix(assistant): first-send auto-create, stream error surfacing, composite delete

ctkm-aelf:
## What broke on prod

Typing into the `/assistant` "New chat" empty state and pressing Send did nothing. Two stacked causes:

1. **FE silent no-op (fixed here):** a fresh account has no conversations → `selectedId` is `undefined` → `useSendMessage` threw internally → `chat-composer` swallowed the rejection and just restored the text. Nothing ever auto-created a conversation on first send; only the "+ New chat" button did. The mock-mode POC never hit this because mock data pre-seeds conversations.
2. **Auth (TD-3, not this PR):** cookie sessions have no Bearer to forward, so live Aevatar 401s every pass-through call. Fix = Aevatar validates `X-NyxID-Identity-Token` + the three-field service-row flip (`identity_jwt_audience:"urn:aevatar:api"`, `inject_delegation_token:true`, `forward_access_token:false`) — per the `eanz17/nyxid-chat` reference README. External dependency; tracked in `docs/CHAT_ASSISTANT_SPECS.md` TD-3 (rewritten in this PR).

## Changes

**Frontend**
- First send auto-creates a conversation and streams into it; the page follows the navigation to the new thread.
- Send failures toast (`describeSendFailure`) — never silent again.
- Pre-SSE error envelopes (`{code,message}` / NyxID `{error,error_code,message}`) surface as the turn error; 401/403 keep the auth-specific "still signed in" copy → **supersedes draft #1195** (same copy, wider coverage), which can be closed.
- EOF without `RUN_FINISHED`/`RUN_ERROR` = failed `stream_closed` turn (truncated streams no longer report success); final unterminated SSE frame flushed; bare-`\r` framing normalized.
- Per-conversation `sessionId` on every `:stream` body (reference-aligned; verified accepted by prod Aevatar via CLI).

**Backend**
- `POST /assistant/conversations`: bounded best-effort wait for the created actor to appear in the actor index (async-202 materialization race; mirrors reference `waitForConversation`).
- `DELETE /assistant/conversations/{id}`: composite dual-delete (actor + chat-history row, 404-tolerant each side, `200 {}`).

## Verification

- Local E2E against a mock Aevatar implementing the nyxid-chat + chat-history contracts: empty-state send → create (202) → materialization poll → stream → rendered reply; composite delete observed as both upstream DELETEs; 401 mode shows the "Message not sent" toast with text restored.
- Full CLI matrix against **prod** stays green (list/create/stream/history/delete), including a stream with the new `sessionId` field.
- Gates: backend `cargo fmt` + `clippy --all-targets` clean + assistant tests 8/8; frontend 1880+ tests, lint 0 errors, `npm run build` (tsc -b) green; wizard bundle byte-identical.

## Not in this PR (tracked in CHAT_ASSISTANT_SPECS.md)

- TD-3 rollout (Aevatar-side identity-token validation, then the row flip — do not flip early, Bearer forwarding is what keeps CLI callers working today).
- TD-7 frame-taxonomy port (`protocol.js`: approval/tool/authorization cards + redaction) — includes fixing `decideApproval` to consume the SSE `:approve` response.
- Startup seed/reconcile of the aevatar row + strict admin-target proxy mode (TD-1/TD-8).

### DIFF
--- a/backend/src/handlers/assistant.rs
+++ b/backend/src/handlers/assistant.rs
@@ -8,25 +8,34 @@
 //! Every route is an explicit mapping onto one Aevatar path. A blanket
 //! `/{*path}` pass-through would expose all ~248 routes of Aevatar's spec
 //! (admin, workflow, and streaming-proxy surfaces included) through a
-//! session-authed mount; the assistant needs eight.
+//! session-authed mount; the assistant needs nine.
 //!
 //! Forwarding goes through `proxy::execute_proxy`, so credential injection,
 //! identity propagation, per-agent rate limiting, approval gating, and audit
 //! all apply unchanged -- the assistant is not a special case in the data
 //! plane (PRD N4). SSE responses stream through it unbuffered.
 
+use std::time::Duration;
+
 use axum::{
-    body::Body,
+    Json,
+    body::{Body, to_bytes},
     extract::{Path, State},
-    http::Request,
-    response::Response,
+    http::{Method, Request, StatusCode},
+    response::{IntoResponse, Response},
 };
 
 use crate::AppState;
-use crate::errors::AppResult;
+use crate::errors::{AppError, AppResult};
 use crate::handlers::proxy::execute_proxy;
 use crate::mw::auth::AuthUser;
 use crate::services::assistant_service;
+
+/// Create responses are a ~300-byte accepted envelope; the cap only guards
+/// against a misbehaving upstream.
+const MAX_CREATE_RESPONSE_BYTES: usize = 64 * 1024;
+/// The actor index is bare `{actorId}` rows (~60 bytes each).
+const MAX_INDEX_RESPONSE_BYTES: usize = 512 * 1024;
 
 /// Resolve the admin-managed Aevatar service and forward `path` to it.
 ///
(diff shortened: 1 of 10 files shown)


ctkm-aelf:
## Adversarial review (codex gpt-5.6-sol) — verdict was FIX-FIRST; all four P1s fixed in c01c5701

| # | Finding | Disposition |
|---|---|---|
| P1-1 | Async stream failures stored but never rendered (mutation resolves at stream start) | **Fixed** — failed `turn.completed` now toasts with the turn error (stable per-turn id). Verified in-browser: mid-conversation 401 shows "The assistant reply failed … could not authenticate you to the chat backend". |
| P1-2 | Synthetic poll/history-delete requests dropped the caller's `Authorization` | **Fixed** — `synthetic_request` helper copies the original header (+unit tests). Verified against the mock in require-bearer mode: the materialization poll and BOTH composite deletes carry the Bearer; delete returns 200. |
| P1-3 | Empty-state auto-create not single-flight | **Fixed** — module-level in-flight promise; racing sends share one conversation, the loser rejects on the active-turn guard. Regression test: `Promise.allSettled` race → exactly 1 `createConversation`, 1 fulfilled, 1 rejected. |
| P1-4 | Reprojection erased the per-conversation `sessionId` | **Fixed** — `loadHistory` + `mergeIndexEntry` preserve it; the cross-turn test now interleaves `getHistory` + `listConversations` between sends. |
| P2-5 | Trailing `\r` eagerly normalized could split a multi-`data:`-line event | **Fixed** — trailing CR held back undecided; regression test with the exact split. (Strictly better than the reference, which has the same eager normalize.) |
| P2-6 | Composite delete short-circuits vs reference attempting both | **Kept, documented** — deliberate: a hard actor-delete failure leaves the conversation intact and retryable instead of deleting the history row and orphaning an invisible actor. Retry converges either way. Rationale now in the handler doc. |
| P2-7 | No handler-level mock-upstream tests | **Partial** — `synthetic_request` unit tests added; full `execute_proxy` mock-upstream harness deferred (covered today by the manual mock-Aevatar E2E; candidate for the TD-7 protocol PR where a shared mock upstream pays for itself). |
| P2-8 | Stale ~1.4s wait-budget docs | **Fixed** — ~2.0s of sleeps (before network time) in both the spec doc and the constant docs; trailing whitespace removed. |

Gates after fixes: clippy clean, backend assistant tests 10/10, FE 1852/1852 + lint 0 errors + build green. E2E re-verified end-to-end on the mock stack (happy path, async-failure toast, Bearer propagation).

PR #1201, opened 2026-07-18
fix(assistant): forward a delegated token so Aevatar's LLM callback succeeds

ctkm-aelf:
## Follow-up to #1200 — the chat still 401'd, one leg deeper

#1200 fixed the chat **entry** (Aevatar accepts the forwarded bearer → `RUN_STARTED`). But a real prod cookie chat then failed with `RUN_ERROR (401)`:

```
Upstream LLM route '/api/v1/proxy/s/chrono-llm-public' rejected the request with HTTP 401 ...
Upstream said: {"error":"unauthorized","error_code":1001,"message":"Unauthorized: Invalid token"}
```

**This is a NyxID self-inflicted bug, fixable entirely NyxID-side.** Aevatar authenticates its own callbacks to NyxID by **reusing the inbound `Authorization` bearer** (the prod row has `inject_delegation_token: false`, so it has no `X-NyxID-Delegation-Token`). When Aevatar called NyxID's LLM gateway on the user's behalf, the cut-4 `assistant_forward` token — which `verify_token` rejects **everywhere** — was refused with NyxID's own `1001 "Invalid token"`. The "reject everywhere" marker never accounted for Aevatar legitimately calling NyxID's proxy/LLM surfaces, which is how chat works.

## Fix: forward a standard **delegated** access token

`assistant::forward` now mints `generate_delegated_access_token` (`delegated: true`, `act.sub = aevatar`, scope `PROXY_SCOPE`, restrictions from `TokenRestrictionClaims::from_auth_user`, TTL `MCP_DELEGATION_TOKEN_TTL_SECS` = 300s) and overwrites `Authorization`. This **is** the documented platform delegation-token standard ("downstream calls NyxID on the user's behalf"), delivered in `Authorization` (the header Aevatar reuses) instead of `X-NyxID-Delegation-Token`.

- Replay boundary moves to the router layer: `reject_delegated_tokens` refuses it on every human-only + shared surface (account / admin / keys → **403**) while `api_v1_delegated` accepts it on `/llm`, `/proxy/s/*`. Strictly safer than the plain access token CLI/Bearer callers already forward and prod already trusts.
- `PROXY_SCOPE` (not the row default `llm:proxy`) is required — the LLM call arrives as a REST proxy passthrough enforcing `ensure_rest_proxy_access`.
- Gate and kill switch unchanged: `Session && forward_access_token`; the TD-3 row flip (`forward_access_token: false` + `inject_delegation_token: true`) retires the Authorization mint and hands over to the standard delegation header, no code change.

## cut-4 marker kept as a one-deploy tombstone

`Claims.assistant_forward`, the `verify_token` rejection, the generator, and the config are **retained this deploy**: a cut-4 token still live at Aevatar (≤300s) must keep being rejected, else serde would decode it as an ordinary access token → reopened replay hole (codex round 5). Marked legacy in config/ENV; generator `#[allow(dead_code)]`. **Follow-up issue will delete the whole marker machinery** once pre-migration tokens expire.

## Verification

- **E2E** against a two-leg mock (Aevatar replays the forwarded bearer to a seeded `chrono-llm-public` → mock LLM): the full run completes (`RUN_FINISHED` with content), the callback leg returns **200**, the same token is **403** at `/users/me` + `/api-keys` + `/connections`, the row flip suppresses the mint, and Bearer callers forward byte-for-byte.
- New wiring-lock unit test asserts scope=`proxy`, `act.sub=aevatar`, `delegated=true`, `assistant_forward=None`, TTL=300s.
- 4682 backend tests + clippy + fmt green.
- Consensus: Claude Fable 5 + codex `gpt-5.6-sol` round 5 (AGREE-WITH-CHANGES, all applied) + codex review (no P1; 2 P2 addressed; final review no findings).

## Residuals (recorded in CHAT_ASSISTANT_SPECS cut 5, not fixed here)

- Delegated callbacks don't inherit `Session`'s approval bypass — a gated write enters approval as requester `delegated`/`aevatar` (correct security; Aevatar must handle the flow).
- `/delegation/refresh` may reject `act.sub=aevatar` (pre-existing standard-family inconsistency) — immediate callbacks work; a run issuing callbacks after 5 min may need a follow-up.


ctkm-aelf:
## Update — architecture audit (codex round 6) + rework applied

Ran an adversarial anti-pattern / BE-architecture audit on the bridge. Verdict: the mechanism (delegated token in `Authorization`) is **NyxID-aligned** — it's a strictly-more-restricted version of what `forward_access_token` already forwards, uses the same `generate_delegated_access_token` factory / TTL / `act.sub=slug` as the standard `inject_delegation_token` path, goes through `execute_proxy` unchanged (no layer bypass), and mints in the handler exactly like the standard path.

**One real anti-pattern was found and fixed:** the earlier cut hardcoded `PROXY_SCOPE`, shadowing `DownstreamService.delegation_token_scope` — the config field the standard path reads as the single source of truth (FI-003/FI-005). That split-brain meant the `Authorization` token and the future `X-NyxID-Delegation-Token` could disagree and silently re-break callbacks at the TD-3 cutover.

Rework (commits `dbbae68c`, `fcf511ff`):
- `build_forward_authorization` now sources scope from `service.delegation_token_scope` and validates `scope_allows_rest_proxy`, **failing closed** (500, detail server-side) on a misconfigured row rather than minting a token that 403s inside Aevatar's callback.
- The `Authorization`-delivered token and the standard `X-NyxID-Delegation-Token` are now **capability-equivalent**; the only remaining deviation is the delivery header (the irreducible Aevatar-compat requirement, already logged as TD-3).
- Regression test proves the exact row scope (`proxy:*`) reaches the JWT.

### ⚠️ Required prod config (new)
The `aevatar` row must set **`delegation_token_scope: "proxy"`** (the default `llm:proxy` cannot reach `/proxy/s/{slug}` and the bridge now rejects it). Valid transitional row state: `forward_access_token: true` + `inject_delegation_token: true` + `delegation_token_scope: "proxy"`. This supersedes the earlier "no DB change needed" note.

Re-verified E2E with the config-driven scope (full run + callback 200) and the fail-closed path (`llm:proxy` row → 500 at NyxID, downstream never called). 4683 tests + clippy + fmt green. Codex round 7: RESOLVED-INLINE.

Issue #1290, opened 2026-07-31
Catalog-backed services with no registered ServiceEndpoint rows are dropped from /api/v1/mcp/config as invalid-contract — leaving every api-lark-bot instance invisible to aevatar v4 workflow admission (user-mounted OpenAPI specs are ignored on the catalog path)


jianwei-su:
### Summary

aevatar's v4 external-capability admission (aevatar #2996, deployed) authorizes workflow
steps strictly against the caller's `/api/v1/mcp/config` catalog. For our Lark-bot service
instance (catalog template `api-lark-bot`, instance owned by us, active, with an OpenAPI
subset spec mounted via `nyxid service update --openapi-spec-url`), the catalog never
publishes it — so every bind attempt fails with
`USER_SERVICE_NOT_VISIBLE: The selected NyxID UserService is not visible to the current caller.`

Reading `backend/src/services/mcp_service.rs`, the mechanism is:

1. **Catalog-backed branch ignores user specs.** For a user service with
   `catalog_service_id` set, endpoints come exclusively from the ServiceEndpoint rows
   "pre-parsed at catalog registration time"; the instance's `openapi_spec_url` is never
   consulted (that path only runs for custom endpoints).
2. **The `api-lark-bot` catalog service has no registered ServiceEndpoint rows**, so the
   instance surfaces with an empty endpoint set.
3. **Empty endpoint set ⇒ dropped.** `operation_set_is_publishable` returns false for
   empty endpoints, and the service is removed and counted in
   `diagnostics.invalid_contract_services` (our caller config currently reports 44).

Net effect: the entire class of catalog-backed credentialed services (Lark bot being our
case) cannot be referenced by aevatar v4 `capability.nyxid_operation` selectors at all.
We verified the two escape hatches are also closed:

- generic-proxy endpoints are rejected by aevatar admission by design
  (`NYXID_GENERIC_PROXY_REJECTED: Generic proxy services are not eligible for workflow admission`);
- draft-runs carry no admission plan, so only a real bind can consume the catalog.

### Impact

Our production-parallel finance workflow (weekly budget monitor: 6 read-only Bitable
reads + 1 message send through the Lark bot instance) is fully rewritten to the v4
canonical form and ready to bind — but cannot pass admission until the Lark bot service
appears in `/api/v1/mcp/config` with real endpoint ids.

### Request

Either (in preference order):

1. **Register parsed endpoints for the `api-lark-bot` catalog service** (at minimum:
[...]

PR #1296, opened 2026-07-31
fix(catalog): strip schema validation keywords rejected by aevatar workflow admission (follow-up to #1290/#1291)

jianwei-su:
## Summary

Follow-up to #1290 / #1291. The overlays shipped in #1291 are live (verified: 45 services in `/api/v1/mcp/config`, `invalid_contract_services` 44 → 12, all `api-lark-bot` instances now publish 5 endpoints) — but binding an aevatar v4 workflow against them still fails, one layer deeper:

```
NYXID_ENDPOINT_SCHEMA_UNSUPPORTED: The endpoint schema is outside the supported workflow contract subset.
```

aevatar's admission proof builder (`NyxIdOperationAdmissionProofBuilder.cs`) accepts only a fixed schema-keyword whitelist — `type, enum, properties, required, items, additionalProperties, title, description, default, example, examples, deprecated` — and drops the whole endpoint on any other keyword (`HasMalformedKeyword`). The #1291 overlays use `minLength` / `minimum` / `maximum`, so 4 of 5 lark-bot endpoints (and equivalents across 17 other overlays: 110 occurrences total) are inadmissible. The #1290 goal — catalog services bindable by aevatar workflows — is therefore still unmet in practice.

## Change

Mechanically strip `minLength` / `minimum` / `maximum` from parameter and body schemas in all 18 affected overlays (78+/81− across single lines; formatting untouched). These keywords are advisory only — the upstream APIs enforce the real limits at runtime — so the change is behavior-neutral for proxying while making every overlay operation admissible.

The alternative (aevatar widening its whitelist) also works but is code + tests + a release on the other side; this is data-only and unblocks immediately. Happy to close this in favor of that route if preferred.

## Verification

- All 22 catalog specs still parse as JSON.
- Whitelist re-scan across all overlays: zero violations remaining except the known residuals below.
- We hold a fully v4-migrated workflow (6 read ops + 1 send op against lark-bot endpoints, opaque endpoint ids) ready to re-bind — we can confirm end-to-end admission within minutes of this deploying.

## Known residuals (left untouched — need semantic decisions, out of scope here)

- `firecrawl.openapi.json`: 3× `$ref` (aevatar rejects unresolved refs)
- `mistral` / `openai` / `telegram-bot`: 4× union `type` arrays (aevatar accepts only single normalized types)

These endpoints remain inadmissible to aevatar until resolved; listing them here so they don't surprise the next consumer.

### DIFF
--- a/backend/specs/catalog/anthropic.openapi.json
+++ b/backend/specs/catalog/anthropic.openapi.json
@@ -33,7 +33,7 @@
                     "description": "Conversation messages with role and content.",
                     "items": { "type": "object", "additionalProperties": true }
                   },
-                  "max_tokens": { "type": "integer", "minimum": 1 },
+                  "max_tokens": { "type": "integer" },
                   "system": { "description": "Optional system prompt." },
                   "temperature": { "type": "number" }
                 }
@@ -69,7 +69,7 @@
             "name": "limit",
             "in": "query",
             "required": false,
-            "schema": { "type": "integer", "minimum": 1, "maximum": 1000 }
+            "schema": { "type": "integer" }
           }
         ],
         "responses": {
--- a/backend/specs/catalog/cohere.openapi.json
+++ b/backend/specs/catalog/cohere.openapi.json
@@ -117,7 +117,7 @@
                   "model": { "type": "string", "description": "e.g. rerank-v3.5." },
                   "query": { "type": "string" },
                   "documents": { "type": "array", "items": { "type": "string" } },
-                  "top_n": { "type": "integer", "minimum": 1 }
+                  "top_n": { "type": "integer" }
                 }
               }
             }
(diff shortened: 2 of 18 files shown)


chronoai-kai:
Folded into #1295 (now merged) — your commit was cherry-picked with authorship preserved, and the rule applied comprehensively: the 61 operations #1295 added had reintroduced the same keyword class, so the whitelist strip now covers all overlays (250 occurrences), and both known residuals are resolved too (firecrawl's 3 $refs inlined, all 11 union type arrays collapsed). A new registry test (`every_schema_stays_inside_aevatar_admission_subset`) fails the build on any non-whitelisted keyword, union type, or $ref, so this can't regress. Thanks for the precise report and fix — your v4 workflow should be bindable once the next deploy rolls out.

Issue #1279, opened 2026-07-30
[Assistant] Route new chats through Aevatar typed NyxIdChat so service.connect cards can render


eanz17:
## Problem

A new chat opened at `https://nyx.chrono-ai.fun/assistant` cannot render the rich `service.connect` card when the requested service is missing.

Reproduction (production, 2026-07-30):

1. Open a new Assistant chat.
2. Send `我要查看 aws 账单`.
3. The UI shows a Workflow run (`reply` / tool steps) and prose explaining that AWS is not connected.
4. No `nyxid.action.request` reaches the browser, so no rich card can appear.

This is a runtime/ownership mismatch, not a card-renderer bug:

- PR #1263 intentionally makes every new conversation a client-local `workflow-pending-*`, aliases it to `chatc-*`, and sends the turn to `POST /api/v1/assistant/workflow-chat`.
- The backend rebuilds that request as Aevatar `POST /api/chat` with `workflow: "studio"`.
- Only pre-existing `nyxid-chat-*` conversations use the typed NyxIdChat stream path.
- The rich-card transport already validates and renders schema-v4 `CUSTOM nyxid.action.request`, but the Workflow Studio path does not produce that NyxIdChat browser-action contract.

## Backend readiness

Aevatar's typed Assistant contract is implemented in current code:

- Mainnet `POST /api/chat` selects NyxIdChat only through an explicit Assistant `type` discriminator and preserves discriminator-free Workflow Chat compatibility.
- First-turn `type: "text"` creates the server-owned NyxIdChat conversation and streams authoritative conversation/turn identity.
- `/api/chat/conversations/**`, typed `action.continue`, approvals, and task/step controls reuse the existing NyxIdChat actor authority.
- The missing-service path commits a blocked turn and emits one schema-v4 `service.connect` action when `nyxid_require_service` returns verified registration-required evidence.
- Aevatar commit [fbd6f9fb3](https://github.com/aevatarAI/aevatar/commit/fbd6f9fb31e72af91441eb6c601ef4eac2fd3ff6) removes legacy human-session management tools such as `nyxid_services` from the typed NyxIdChat surface.

The remaining product mismatch is in NyxID: `main` still creates `workflow-pending-*`, resolves it to `chatc-*`, and submits new Assistant turns through `/api/v1/assistant/workflow-chat`.

[...]

PR #1301, opened 2026-07-31
fix(assistant): route new chats through the studio workflow engine

ctkm-aelf:
## Problem

New chats were **not** running on Workflow Studio. The upstream `POST /api/chat` body carried no `workflow` field:

```json
{"type":"text","prompt":"i want to connect to github","clientRequestId":"acc32ecc-…"}
```

Aevatar's `/api/chat` is a dispatcher that branches on the **presence** of `type` (`MainnetChatEndpoints.ClassifyRequestAsync`):

| Body | Kind | Handler |
|---|---|---|
| has `type` in the assistant allowlist | `Assistant` | `NyxIdChatEndpoints.HandlePublicChatAsync` |
| **no** `type` | `Workflow` | `WorkflowCapabilityEndpoints.HandleChatPostAsync` — the only branch that reads `workflow` |
| unknown `type` / non-JSON | `Unsupported` | 400 `INVALID_CHAT_INPUT` |

The two surfaces are mutually exclusive per turn, so a `type`-bearing body can never reach the studio engine — adding `workflow` to it would just 400 (`PublicStreamRequest` is `[JsonUnmappedMemberHandling(Disallow)]` and has no `Workflow` member).

Root cause: `createConversation()` minted a `nyxid-pending-` placeholder, which selected the `typed-create` protocol introduced in #1279.

## Fix

Mint the workflow placeholder instead, so a new conversation's first turn posts to `/api/v1/assistant/workflow-chat`. **No backend change** — `assistant_service::workflow_chat_body` already pins the workflow server-side and produces the required payloads:

```json
// first turn
{"commandId":"…","conversation":{"conversationId":null},"prompt":"…","workflow":"studio"}
// follow-up
{"commandId":"…","conversation":{"conversationId":"chatc-…","minimumStateVersion":3},"prompt":"…","workflow":"studio"}
```

The now-unreachable `typed-create` run protocol and its `RUN_STARTED` actor-adoption block are removed (FI-007).

## What is preserved

- **Existing `nyxid-chat-…` conversations** keep the typed actor surface, so transcripts created while it was live stay continuable.
- **Action cards** still render — `handleCustomFrame` parses `nyxid.action.request` independently of protocol.
- **`action.continue` / `approval.resolve`** stay on the actor protocol, addressed by the actor id the action frame names, never the `chatc-…` id (`drainPendingActions` already enforced this and blocks the batch with an outcome note when the frame names no actor).
- The `nyxid-pending-` prefix survives only in the `getHistory` not-found guard, so a stale `?c=` URL from a pre-cutover session still resolves cleanly.

## Known limitation (pre-existing, unchanged)

`decideApproval` throws for workflow conversations — the workflow surface resumes via `runs/{runId}:resume`, which this mount does not proxy. Studio chats therefore cannot decide an approval inline; the message directs the user to NyxID → Approvals. This was already true for legacy `chatc-…` rows and now applies to new chats too.

## Verification

- `npm run build` (tsc -b + vite build, the CI gate) — clean
- `npm run lint` — 0 errors
- Full frontend suite: **193 files / 2273 tests passing**
- Assistant transport suite rewritten for studio-first routing (132 tests), asserting the first turn posts `{prompt, commandId}` **only** — no `type`, no `conversationId` — to `/workflow-chat`

### DIFF
--- a/backend/src/services/assistant_service.rs
+++ b/backend/src/services/assistant_service.rs
@@ -1075,6 +1075,18 @@ pub struct WorkflowChatTurnRequest {
     /// same conversation/turn for a repeated id, 409s on payload mismatch).
     #[serde(default)]
     pub command_id: Option<String>,
+    /// Client-controlled session correlation handle, stable for the life of
+    /// one conversation. Aevatar's `HttpChatInput.SessionId` is optional and
+    /// falls back to the run's correlation id
+    /// (`WorkflowChatRequestEnvelopeFactory`), so this is correlation
+    /// plumbing, not conversation identity — chat continuity is carried by
+    /// `conversation.conversationId` + `minimumStateVersion`.
+    ///
+    /// It IS part of Aevatar's create-replay fingerprint
+    /// (`WorkflowChatCreateRequestFingerprint`), so a retry of the same
+    /// `commandId` MUST repeat the same value or the replay 409s.
+    #[serde(default)]
+    pub session_id: Option<String>,
 }
 
 /// Opaque client token (command ids): UUID-shaped material only, so nothing
@@ -1139,12 +1151,22 @@ pub fn workflow_chat_body(request: &WorkflowChatTurnRequest) -> AppResult<serde_
         None => uuid::Uuid::new_v4().to_string(),
     };
 
-    Ok(serde_json::json!({
+    let mut body = serde_json::json!({
         "commandId": command_id,
         "conversation": conversation,
         "prompt": request.prompt,
         "workflow": WORKFLOW_CHAT_WORKFLOW,
-    }))
+    });
+
+    // Omitted rather than sent null when the caller has none: Aevatar
+    // normalizes blank to null anyway, and an absent member keeps the body
+    // identical to the pre-sessionId shape for callers that never send one.
+    if let Some(session_id) = &request.session_id {
+        validate_client_token(session_id, "sessionId")?;
+        body["sessionId"] = serde_json::Value::String(session_id.clone());
+    }
+
+    Ok(body)
 }
 
 /// `api/ws/chat` -- WebSocket twin of the workflow chat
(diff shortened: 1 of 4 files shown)


ctkm-aelf:
## ⚠️ Confirmed trade-off — read before merging

Routing new chats to the studio workflow engine means **both NyxID-specific card surfaces stop rendering**. Verified against `aevatarAI/aevatar` `origin/dev`, not inferred:

| Card | Frame | Producer | On studio? |
|---|---|---|---|
| Action card (`service.connect`) | `nyxid.action.request` | `agents/Aevatar.GAgents.NyxidChat/NyxIdChatConversationAguiFrameBuilder.cs` | **Lost** |
| Connect card | `nyxid.authorization.required` | `agents/Aevatar.GAgents.NyxidChat/*` (4 sites, all in that agent) | **Lost** |
| Approval card | `aevatar.tool_approval.pending` | `src/workflow/Aevatar.Workflow.Presentation.AGUIAdapter/EventEnvelopeToWorkflowRunEventMapper.cs:700` | Renders, but `decideApproval` throws for workflow conversations → dead card |

`git grep -l "NyxIdChat" src/workflow/**/*.cs` returns **nothing**: there is no workflow→NyxIdChat bridge. Since `MainnetChatEndpoints.ClassifyRequestAsync` dispatches on the presence of `type`, a studio body (no `type`) never reaches the agent that emits those two frames.

This is the exact capability #1279 was built to enable ("Route new chats through Aevatar typed NyxIdChat so service.connect cards can render"). The two requirements are in direct conflict **upstream** — no NyxID-side change resolves it. Options:

1. Merge as-is: studio engine, no action/connect cards.
2. Stay on the typed surface: cards work, no `workflow: "studio"`.
3. Ask Aevatar to emit `nyxid.action.request` / `nyxid.authorization.required` on the workflow run path, then merge this.

## Also not fixed here

Stopping a studio turn is client-side only — `requestServerStop` returns `null` for workflow runs, so the upstream run keeps executing while the UI reports "cancelled". Pre-existing for legacy `chatc-…` rows; this change makes it apply to every new chat. The real stop is `runs/{runId}:stop`, which this mount does not proxy.

## Fixed in this PR (from adversarial review, gpt-5.6-sol)

- A create turn whose context frame named no valid `chatc-…` completed *successfully* while the record stayed on its placeholder — the next send then minted a **second** upstream conversation. Now fails closed.
- A replay returning a different `chatc-…` or turn id silently re-keyed the conversation while attributing later events to the old identity. Now fails closed.

Both were guards the deleted `typed-create` block had, which I removed without porting a workflow equivalent.

PR #1352, opened 2026-08-05
fix(oauth): retheme the MCP authorize flow to the current design system

ctkm-aelf:
Cherry-picks [`6c5c1ed6`](https://github.com/ChronoAIProject/NyxID/commit/6c5c1ed69240c2e6d0a111477f69627d2c4acb33) (originally #1351, landed on `rollup-chat-2026-08-04`) directly onto `main`. **The rollup branch is abandoned** — this PR is the delivery path for that work.

The cherry-pick applied cleanly and the resulting diff is byte-identical to the original commit.

## What it does

The consent screen — what a user sees when authorizing the NyxID MCP server from Claude Code, Cursor, or claude.ai — rendered a white card on a black void in light mode, and used a layout no other page in the app shares.

`index.html` sets an inline `background-color:#0D0D0D` on `<body>` as an anti-flash guard for the never-themed public pages. Inline style beats the `bg-background` class, so a themed standalone page that doesn't paint its own canvas shows black gutters. Every other themed surface paints one; the consent page was the only one that didn't.

The page now mirrors `login-device.tsx`, its closest sibling (the other standalone "a human reviews and approves an access request" surface): same shell, h1 header outside the Card, shared DetailSection/DetailRow blocks, shared ErrorBanner, semantic warning callout, responsive action row.

Shared components:
- `DetailSection` gains optional `action` + `className`; `DetailRow` gains `mono` plus min-w-0/break-words/shrink-0 so long IDs and URIs wrap rather than overflow.
- `button.tsx` and `auth-flow.tsx` migrate `white/[α]` chrome to the hairline / overlay tokens (app.css "Interactive chrome"). The token dark values reproduce the originals 1:1, so dark mode is pixel-unchanged; light mode gains the borders and hover states it was missing.
- Risk pills move from raw `red-*`/`yellow-*`/`emerald-*` to shared Badge variants per DESIGN.md (high→destructive, medium→warning, low→success). `scopeRiskClass` is deleted and both callers migrated.

Backend success page (loopback and custom-scheme redirects) moves onto the app's surfaces, the real SuccessPanel treatment, the correct nyx-gradient-vivid direction, and an inlined vector wordmark instead of typed text.

## Extra commit vs. the original

`button.tsx` is a manifest-listed source of the CLI wizard bundle, so the token migration made `cli/src/wizard/bundle-meta/index.hash` stale. `cargo test -p nyxid-cli --test wizard_bundle_freshness` was **red** on the original commit — rollup PRs get no CI (`ci.yml` gates `main`/`dev` only), so it was never caught there. Second commit rebuilds the bundle on this base. `index.manifest` is unchanged (no imports added or removed); the `index.html` delta is exactly the button token migration plus the Tailwind utilities dropped along with `scopeRiskClass`.

## Verification (local, full CI-equivalent)

| Check | Result |
| --- | --- |
| `cargo nextest run -p nyxid` | 4984 passed |
| `cargo nextest run -p nyxid-cli` | 1079 passed |
| `cargo test -p nyxid-cli --test wizard_bundle_freshness` | green (red before the rebuild commit) |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |
| `cargo fmt --all -- --check` | clean |
| `npm run test` (frontend) | 2601 passed / 214 files |
| `npm run lint` (frontend) | 0 errors (23 pre-existing warnings) |
| `npm run build` (frontend) | passes |

## Known follow-up (carried over from the original)

The backend success page is still a hand-maintained replica of the design system in Rust string literals and will drift on the next token change; `credential-accept.html` has the same problem and is off-theme today. Both want the wizard-bundle treatment (built artifact + `rust_embed`).


ctkm-aelf:
Rebased onto `main` after #1349 landed.

#1349 also rebuilt the CLI wizard bundle, so the two branches conflicted on the generated artifacts (`cli/src/wizard/assets/index.html` + `bundle-meta/index.hash`) — a regeneration conflict, not a semantic one. **Zero source-file overlap** between the two commits.

Resolved by dropping the generated commit, rebasing the source cherry-pick onto the new `main`, and rebuilding the bundle on that base — so the hash now closes over both #1349's changes and this one rather than reverting main's. The cherry-picked diff is still byte-identical to `6c5c1ed6`; `index.manifest` is still unchanged.

Re-verified everything on the new base:

| Check | Result |
| --- | --- |
| `cargo nextest run -p nyxid` | 4997 passed |
| `cargo nextest run -p nyxid-cli` | 1079 passed |
| `wizard_bundle_freshness` | green |
| `cargo clippy --workspace --all-targets -- -D warnings` | exit 0, clean |
| `cargo fmt --all -- --check` | clean |
| frontend `npm run test` | 2592 passed / 217 files |
| frontend `npm run lint` | 0 errors |
| frontend `npm run build` | exit 0 |

PR #1382, opened 2026-08-06
fix(assistant): decode Aevatar workflow frames to their real contract

ctkm-aelf:
## What broke

Every **new** chat's first turn failed with "The assistant stream returned an invalid state version." Continuations were fine. The reply was generated and persisted upstream — it appeared on reload — so the client was discarding a good turn.

Aevatar serializes chat SSE frames with `WithFormatDefaultValues(false)`, so a proto3 field holding its type default is omitted from the JSON entirely. `WorkflowChatContextPayload.state_version` is the conversation's **pre-turn** watermark:

```csharp
conversationResolution.ConversationContext?.StateVersion ?? 0   // create → 0
StateVersion = Math.Max(0, context.StateVersion)                // always populated
```

A create turn's value is 0, so the member never reaches the wire — verified against production, where `stateVersion` appears 30+ times as `1,2,3,…,22` and **never once as `"0"`**. `Number(undefined)` is `NaN`, and the guard rejected it as malformed.

Aevatar is conformant here. The strict check came from aligning with Aevatar's console, which uses the *same predicate* but returns `null` to skip an optional extraction — NyxID promoted it to a fatal `deliveryProtocolError`. Every workflow fixture supplied a `stateVersion`, so all of them modelled continuations and none caught it.

## Also fixed

Auditing the rest of the response surface found the same class, plus handlers still pointed at the actor protocol that chat left when turns moved to the studio engine (`#1263`):

| Issue | Cause |
|---|---|
| Failed steps rendered **"Completed"** | `aevatar.step.completed.success` is a proto3 bool; a failure omits it, and `success !== false` inverted it |
| **Connect cards could not appear** | Studio emits no `nyxid.authorization.required`; the verdict rides inside the `nyxid_require_service` result as snake-case JSON |
| **Human-input suspensions vanished** | Cast to `ToolApprovalPayload`, whose identity guard dropped it — that payload has no request id at all |
| **Legitimate waits killed at 120s** | `waiting_signal` left the progress watchdog armed → `upstream_progress_timeout` |
| **Media attachments dropped** | NyxID read a flat `mediaContent` envelope arm the workflow proto doesn't define; fields are nested under `part`, and the location is `uri` |

Human input renders as prose, deliberately **not** an approval card: the payload carries no approval id, so an approve/deny control would post a decision against an identity upstream never issued. A real input control needs an upstream resume identity first.

## Structural change

Adds `protoBool` / `protoString` / `chatContextStateVersion` so proto3 presence is expressed once rather than re-derived at each call site — that re-derivation is why this class produced two independent bugs. Every change is mirrored in the replay projector so wire-log parity holds.

## Verification

- Regression coverage includes a **sanitized capture of a real production create turn** (`__fixtures__/aevatar-workflow-create-stream.sse`) replayed through the live transport — the shape no synthetic fixture modelled.
- All 12 new tests confirmed to fail without the fix and pass with it, plus a control test proving the watchdog still fires when a run genuinely goes silent.
- `npm run build` clean, lint 0 errors, **2648/2648 tests pass**.
- Backend relay verified lossless — all of this was frontend decoding.

## Follow-ups (not in this PR)

- `projectionStatus` is shipped by Aevatar and referenced nowhere in NyxID — it's the materialization signal requested during the new-chat 404 work.
- Ask Aevatar to document `state_version` as the pre-turn watermark; their design doc still describes the payload as three fields.
- Their console has the same latent `stateVersion` predicate.


ctkm-aelf:
## Review round 1 (Codex Sol) — addressed in e2c3ab38

Verdict was "do not merge as-is" with six findings. I verified each against `origin/dev` before acting. Five were real and are fixed; one I'm handling differently; one it retracted itself.

**Fixed**

1. **`signal.buffered` never re-armed the watchdog.** Real bug I introduced. The per-frame `armWatchdog` runs *before* dispatch while the run is still flagged suspended, so it cleared the timer without setting one — a run stalling right after its signal landed hung until the user pressed Stop. Now re-arms in the handler after clearing the flag. New test pins it.

2. **`parseToolResultBlocker` under-validated.** `blocked: true` + a slug was enough, so any tool emitting those two keys could mint a bogus connect card and fail its own step. Now requires the full readiness DTO (`ServiceRegistrationRequired` + non-empty reason code + non-empty message) — upstream's own invariant for setting `blocked`. Slug regex widened to `[a-z0-9._-]` to match upstream normalization, closing a reverse miss. Four adversarial cases added.

3. **`part.uri` → unsanitized `href`.** Confirmed: `download_url` is rendered directly as an anchor href, so a `javascript:` URI was one click away. The flat arm always had this hole; making the nested workflow shape reachable made it exploitable. Scheme allowlist added; non-http(s) degrades to the shapeless-media notice.

4. **Replay parity gaps.** The completion-receipt path still scored `AUTHORIZATION_REQUIRED` a success and never parsed blockers; human input didn't move its step to waiting; there was no `awaitingSignal` equivalent, so a capture legitimately *ending* at a suspension was scored a truncated failure. All three fixed.

5. **False claim in a comment.** Sol was right — a resume identity does exist (`CanResume` is `runId + stepId`, both on the payload). Comment corrected.

**Handled differently**

- **Human input is "cosmetic".** Fair, and the comment is now honest about it. But an input control needs a NyxID resume route plus a card type — feature work, not frame decoding. What ships here is strictly better than before: previously the ask was dropped *and* the watchdog killed the turn at 120s with a misleading timeout. Now the question is visible and the run waits, matching the precedent the approval gate already sets. Tracking the control as follow-up.

**Retracted by the reviewer**

- The suspected connect-card text divergence between live and replay isn't real — the adapter's strings pass through `safeErrorMessage` unchanged.

**On its test critique** — correct, and my PR wording was imprecise. 12 tests fail without the fix; the others are deliberate controls and negatives that must pass in both states (that's their job — the watchdog control exists so the suspension test can't pass vacuously). Also fixed: the media fixture now uses protobuf-JSON's real enum member name, and the SSE fixture's trailing blank line is gone (`git diff --check` clean).

Gate after changes: build clean, lint 0 errors, **2653/2653 pass**.

PR #1391, opened 2026-08-06
fix(oauth): repair chat popup completion routing, card settlement, and completion copy

ctkm-aelf:
## What was broken

The chat connect-card OAuth popup (PR #1349) was dead on arrival in prod. The backend redirects the provider callback to `{FRONTEND_URL}/oauth?status=…&flow=cc&nonce=…`, and on `https://nyx.chrono-ai.fun`:

```
GET /oauth?status=complete&flow=cc&nonce=…   → 301  location: /oauth/?status=…
GET /oauth/?status=complete&flow=cc&nonce=…  → 404, content-length: 0
```

The 404 is the **Rust backend**, not the SPA (byte-identical headers to `GET /api/v1/<nonexistent>`: doubled security headers, `content-security-policy: default-src 'none'`). Root cause is in `frontend/nginx.conf.template`: `location /oauth/ { proxy_pass … }` makes nginx auto-301 the slashless `/oauth` into the proxied prefix (documented nginx behavior for proxied slash-terminated locations), which lands on the backend IdP router that has no root `/oauth` route. The original implementation plan's Correction C2 predicted the exact opposite ("bare `/oauth` … falls through to the SPA `try_files`") — which is why the completion page was stranded at `/oauth` while the interstitial was correctly moved to `/oauth-launching`. The popup therefore rendered a blank 404, never opened `BroadcastChannel("nyxid.oauth.<nonce>")`, never posted `oauth_result`, and the chat tab got no signal at all.

On top of that, the two owner requirements were unmet: the connect card had no way to learn that an authorization finished (success, denial, or abandonment), and the completion page's copy said nothing about what happened.

## What changed

**Route move + rollout (the P1):**
- Completion page moved to `/oauth-complete` (router, `history.replaceState` scrub, `isPublicPath`, telemetry sensitive-path drop, backend `redirect_to_oauth_completion`, tests). `/oauth/*` stays exclusively the backend IdP surface.
- `frontend/nginx.conf.template`: temporary exact `location = /oauth { return 302 /oauth-complete$is_args$args; }` compat alias (302 on purpose — must not cache durably; removal tracked in `OAUTH_CHAT_POPUP_STATUS.md` §6), plus `Cache-Control: no-cache` on SPA HTML per `docs/DEPLOYMENT.md` (previously unimplemented; security headers re-declared because location-level `add_header` discards inherited ones).
- New `router.test.ts` pins the registered route path — reverting it now fails the suite.

**Requirement (a) — "card knows the state when auth is connected":**
- `AddKeyDialog` announces a structured attempt (`{keyId, attemptId, previousAuthorizationAt}`) via `onAuthorizationPending` only once a real provider journey exists, and fires a new `onAuthorizationAborted` when an announced attempt dies without a server verdict (initiate failure, dialog close = explicit cancellation).
- `useKeyAuthorizationWatch` / `useKeyAuthorizationStatus` are now attempt-generation-isolated (query key `["keys", keyId, "authorization", attemptId]` — a retried attempt can never settle from the previous attempt's cached terminal row) and reconnect-baseline-aware (`active` is terminal only when `last_authorized_at` advanced past the pre-attempt baseline; `undefined` baseline = fresh add).
- The connect card watches the placeholder key via authenticated `GET /keys/{id}` — presence-gated, deadline-bounded (`connectWatchDeadline`, ≤30 min ceiling), focus-refetching — and renders distinct **Connected / Failed (server message) / Timed out / Cancelled (neutral, not destructive)** states with recovery actions, with no user click required. The popup broadcast remains a wakeup optimization only; server truth is the only verdict path.

**Requirement (b) — "redirect page is indicative of what has happened":**
- The page names the service (validated, length-capped label carried through the trusted interstitial handshake into tab-local sessionStorage — never from the query string), maps every emitted backend error code to specific copy, gives the no-peer case honest guidance, keeps the query scrub, and gates all enrichment to `flow=cc` + a nonce-matched launch context. Reserved flows keep byte-identical behavior.
- Popup handoff failures (cold interstitial, blocked window, pre-nonce backend during rollout) degrade to the validated manual link without destroying the attempt; the ready budget is 5s and expiry is non-destructive.

## The security decision, stated plainly

**The success page never claims the connection succeeded.** NyxID hands the provider the attempt nonce as OAuth `state`, and the provider controls where the popup navigates next — so a provider (malicious or compromised) can drive the popup to `/oauth-complete?status=complete&flow=cc&nonce=<real nonce>` with no NyxID callback ever having run. Any client-side "connected" claim on this page is therefore forgeable, and nonce-matching cannot fix that. What the page does assert is only what it can know: an authorization response arrived, which service this window was opened for (tab-local trusted label), and where the authoritative answer lives — the chat card, which settles exclusively from authenticated server reads. The adversarial reviewer constructed this exact forged-provider attack against the shipped page and could not elicit an outcome claim. A popup that *can* say "connected" requires the backend-signed completion receipt specced as a follow-up in `OAUTH_POPUP_FLOW_PLAN.md` §10.1.

## Review trail

This branch went through a full adversarial pipeline; the documents are committed in `docs/`:

1. Plan v1 → **adversarial plan review: REWORK** (`OAUTH_POPUP_COMPLETION_FIX_SOL_REVIEW.md`, 8 findings — all accepted, including withdrawing the plan's original nonce-gated "authorized" success copy as forgeable).
2. Plan v2 (`OAUTH_POPUP_COMPLETION_FIX_PLAN.md`) → implementation round 1.
3. **Adversarial code review: REWORK with two P1 bugs** (`OAUTH_POPUP_COMPLETION_FIX_OPUS_REVIEW.md`):
   - **P1-1:** the card treated `UserService.is_active` as authorization evidence. `GET /keys` only returns `is_active: true` rows, so cancelling a fresh connect rendered a durable **"Connected"** badge over a still-`pending_auth` placeholder.
   - **P1-2:** the same defect in `connectedNow` made the entire new state machine (Authorizing/Failed/Timed out/Cancelled) unreachable on `NYXID_SERVICE_NOT_CONNECTED` cards — a bare placeholder row already read as Connected. A false test fixture (`is_active: false`, impossible server shape) had hidden both.
4. Round 2 fixed both at the root (`UserApiKey.status === "active"` everywhere, plus timestamp advancement for reconnects), corrected the fixtures, restored the manual-link degrade on popup handoff failure, and added the missing route/generation coverage. The reviewer independently re-ran every revert-proof (each fix's test fails when its production line is reverted) → **APPROVE, no blockers**.
5. **Round 3 — rebase-resolution review → APPROVE WITH CHANGES (one P2), both items fixed.** After the post-approval rebase (next section), the reviewer re-reviewed the resolution in isolation and found:
   - **R-1 (P2):** the deliberate `ActionCard` / `ConnectCard` asymmetry — action cards survive dialog close, connect cards treat it as cancellation — was load-bearing but unpinned: wiring `onAuthorizationAborted` into `ActionCard` left its suite green (the test mock never fired the abort), so a future contributor "fixing the inconsistency" would silently restore #1384's stranded-`Connecting` bug with green CI. Fixed: the invariant is named in a call-site comment, the mock now fires the abort on dismiss, and a regression test proves the pending record and "Waiting for authorization" state survive it — revert-proven (wiring the prop in fails the new test).
   - **R-2 (P3):** `attemptId` absence was silently bimodal. Fixed with a fail-safe settlement identity (`pendingAuth?.attemptId ?? \`missing:${block.block_id}\``) used consistently by the query generation and both settle guards, so a malformed record stays settleable instead of stranding the card, block-scoped so malformed records can't share one generic generation across cards. Mutation-tested.

Three P3 observations from the round-2 approval are addressed in a post-approval polish commit (`fix(oauth): require absolute manual-fallback authorization URLs`): the fallback validator no longer base-resolves junk to same-origin links, and the two implicit traps (empty-vs-absent `attempt_nonce`; popup-only href normalization) are documented in place.

## Post-approval rebase — reviewed in round 3

`main` moved ~60 commits while this branch was in review (notably #1384, which moved the action card's `pendingAuth` into the new `pending-connect-store`, and #1356, plugins-page popup). The rebase onto `047dc0b4` had conflicts only in `action-card.tsx`/`.test.tsx` (plus the generated wizard hash). Resolution: kept main's store-based `pendingAuth` (its remount-survival design) and extended `PendingConnectAuthorization` with this branch's `attemptId` + `previousAuthorizationAt`; settle guards now key on `attemptId`; the ActionCard deliberately does **not** pass `onAuthorizationAborted`, preserving main's "dialog close is not abandonment" semantics for action cards (the chat connect card opts into explicit cancellation) — that asymmetry is now test-pinned per R-1.

The round-3 review bounded how much of the rebase was genuinely new surface: of the 26 source files this branch touches, 19 per-file contributions are byte-identical before and after the rebase, and **14 files' final content is byte-identical to the round-2-approved tip** (including `nginx.conf.template`, both popup pages, `connect-card.tsx` + test, `use-keys.ts`, and the whole lib/schema/type surface); `backend/src/handlers/user_tokens.rs` differs only by context drift around the same one production line and two test asserts. The genuinely new surface — the action-card/store merge and the two round-3 fixes — is what round 3 reviewed and approved.

## Gates

Round-3 re-run at `622946d1` (frontend-only changes since the rebase gates): build pass, lint 0 errors, full frontend suite **2759/2759** at `--maxWorkers=2`, wizard rebuild byte-identical (no generated commit needed), freshness green. The two commits after that are a behavior-free rename (CodeQL classification, see note below) and an empty CI retrigger; PR CI validates the final head. No backend code touched after the rebase gates below.

Post-rebase full gates (at the rebase commits):

- `npm --prefix frontend run build` — pass (`tsc -b` + both Vite builds + mock-footprint assertion)
- `npm --prefix frontend run lint` — pass (0 errors, 23 pre-existing warnings)
- Focused suites on every touched surface — 116/116
- `npm --prefix frontend test -- --maxWorkers=2` — **2757/2757 passed (235 files, 413s)** on the clean run. Honest note: this host needs `--maxWorkers=2` — default concurrency (12 cores, load average ~193 during the review runs) produces unrelated 5s-timeout flakes in files this branch never touches; the adversarial reviewer verified the failing set changes run-to-run while every touched suite stayed green in every run. My first post-rebase full run also had a single failure (2756/2757) whose identity was lost to log truncation; the immediate re-run above and the focused touched suites were fully green.
- `cargo test -p nyxid` — **5115/5115 passed** (484s; the suite grew from the review's 4997 with `main`'s new tests)
- `cargo test -p nyxid-cli --test wizard_bundle_freshness` — pass (bundle rebuilt after the rebase, as the last source-affecting commit)

## Explicitly outstanding — this PR approves the code, not the deploy

- **Frontend-image-first rollout is a hard requirement.** The old JS bundle renders a dead not-found page for `/oauth-complete`; the nginx alias makes old-backend redirects work the moment the frontend image ships, so frontend → verify → backend is the zero-broken-window order.
- The plan §3.1d smoke checks (execute the page, assert "Authorization declined" copy + query scrub on both `/oauth-complete?...` and the `/oauth` alias) are still prose, not a checked-in Playwright spec. Run them against the deploy target.
- The §7 manual Chrome/Firefox/Safari (+ one mobile) popup matrix from `OAUTH_CHAT_POPUP_STATUS.md`, now including the new completion/cancellation states.
- Deferred by design: backend-signed completion receipts (`OAUTH_POPUP_FLOW_PLAN.md` §10.1); removal of the temporary nginx `/oauth` alias after one release cycle; remaining F-8 nits from the review (scoping SPA `no-cache` away from `/docs/*` assets, "View connection" button label, duplicate `Cache-Control` on hashed assets, stale route references in the two historical popup docs).
- On a `NYXID_UNAUTHORIZED` card, automatic settlement requires an attempt announced through that card's own dialog: reconnecting entirely elsewhere (e.g. `/keys/{id}` in another tab) leaves the card showing "Reauthorization required" until a new turn re-renders the block. That is the static block contract's boundary, not a regression — without an announced baseline the card cannot distinguish a fresh authorization from an old one.

**CodeQL note:** the first analysis raised one alert — `js/clear-text-storage-of-sensitive-data` on the popup launch-context sessionStorage write (`lib/oauth-popup.ts`). The stored record is `{providerOrigin, correlationId, serviceName}`: the correlation id is the attempt nonce, which is a documented **non-secret** (status doc §4.2 — it transits the provider as OAuth `state` and appears in the completion URL by design), and the provider origin was already sessionStorage-stored by shipped #1349. Addressed the same way as the earlier CQ-1 finding — by making the code's semantics analyzer-visible (`fix(oauth): classify popup launch data as metadata`: pure rename, `nonce`→`correlationId`, zero behavior change, every validation intact) rather than by dismissing the alert. Separately, expect possible false positives against the minified `cli/src/wizard/index.html` (known generated-bundle pattern) — those are dismissed via API individually, never by patching the generated file.

**Outage note:** the 2026-08-06 GitHub Actions major outage swallowed this PR's original `pull_request` workflow runs and failed its CodeQL analyses with runner-acquisition timeouts (`main`'s own release gate at `047dc0b4` went red the same way). The final empty commit exists solely to re-trigger CI/CodeQL cleanly after recovery — default-setup CodeQL runs cannot be rerun via the API.


ctkm-aelf:
## ⚠️ Deploy ordering constraint — read before promoting images

This change is **not purely additive**. The completion surface moves `/oauth` → `/oauth-complete` (frontend route + backend redirect literal), made one-way-safe by a temporary nginx alias. **Merging is inert** — `Publish Images` tags `main`/`edge`/`main-<sha>` and never touches `latest`, and prod compose pins `latest` — but promotion must be ordered.

### Required order: frontend image FIRST, then backend

| Combination | Result |
|---|---|
| New FE + old BE | ✅ nginx alias 302s `/oauth?…` → `/oauth-complete?…`, query byte-exact |
| **Old FE + new BE** | ⚠️ backend redirects to a route the old SPA lacks → `AppNotFound`, no completion broadcast |
| New FE + new BE | ✅ |

### Gotcha: `docker compose up -d` does the WRONG order

`docker-compose.prod.yml:59` declares `frontend depends_on backend`, so a plain `up -d` recreates the **backend first**. Use explicit no-deps ordering:

```bash
docker compose -f docker-compose.prod.yml up -d --no-deps frontend
# verify /oauth-complete renders and /oauth 302s with query preserved
docker compose -f docker-compose.prod.yml up -d --no-deps backend
```

### Blast radius if mis-ordered
Bounded and self-healing. The credential is written **before** the redirect, so nothing is lost — the popup shows a dead page and the opener sits on "Connecting" until polling finds the active credential or times out. Note prod's *current* behaviour for this flow is already a dead popup (`301 → /oauth/` → zero-byte 404), so the skew window swaps one dead end for another rather than regressing.

### Also in-flight (deploy-window only, low severity)
sessionStorage moved from scalar `nyxid.oauth.provider-origin` to JSON `nyxid.oauth.launch-context` with no dual-read. A popup launched pre-deploy still broadcasts its result, but the new page can't recover the old provider origin — inline retry is disabled and that user restarts the attempt. Verified to degrade gracefully (no throw) in both directions.

### Not required
No provider console reconfiguration — redirect URIs point at `{BASE_URL}/api/v1/providers/callback`, not the frontend page. Wire protocol untouched (BroadcastChannel name, `state=1cc_<uuid>`, `?nonce=` all unchanged).

### Still manual after deploy
Frontend-image-first verification, the §3.1d smoke (still prose, not a checked-in spec), and the §7 Chrome/Firefox/Safari popup matrix.

Issue #1457, opened 2026-08-17
[Assistant][Delegation] Close post-#1449 provider-identity and cross-flow acceptance gaps


eanz17:
## Context

Follow-up to #1439, #1440, and #1449. The implementation from #1451 / #1453 is present on NyxID `main`, and production currently reports `v0.10.0 / ebce858964af`.

Aevatar Milestone 42 consumers (aevatarAI/aevatar#3377, #3364, #3369, and #3366) have completed their local implementation, but production delivery remains fail-closed until the NyxID producer contract is fully evidenced. Closing #1449 does not by itself satisfy that gate.

The latest automated evidence recorded on #1449 still says:

- `gate_evidence_manifest=pending`
- `gate_post_merge_probe=pending`
- `post_merge_probe_green=pending`

No open successor issue currently carries the remaining questions below.

### 1. The additive exact-view digest does not visibly bind canonical provider identity

`ExactOperationViewService` currently projects `user_service_id`, `service_slug`, node information, and operations, but no immutable catalog-provider identity (see `backend/src/services/mcp_service.rs`, around the `ExactOperationViewService` definition and `exact_operation_view` projection).

For user-managed services, the projected slug is populated from the proxy-side `UserService.slug`. The source `catalog_service_id` is used while loading the catalog but is not retained in the final caller-visible exact view or its digest.

Please either:

1. carry a stable canonical provider identity (for example the catalog service ID or another producer-owned immutable key) into the shared exact projection and `exact_view_digest`; or
2. provide code-verifiable evidence that the current slug is unique, immutable, cryptographically bound to the actual catalog provider, and cannot drift independently of that provider.

A mutable proxy/UserService slug alone is not sufficient evidence of provider identity.

### 2. Discovery -> approval create/redeem is not covered as one real flow

Current handler tests exercise catalog discovery, while exact-approval tests exercise create/redeem with separately or manually constructed digest/binding fixtures. They do not demonstrate that a response produced by the real delegated discovery handler is accepted by create/redeem and then rejected after relevant drift or revocation.

Add an integration test that:

1. issues a valid owner/client/service/node-bound delegated token with `mcp:catalog:read`;
2. obtains the actual `GET /api/v1/delegation/operation-catalog` response;
3. uses that exact response/binding for delegated exact-approval creation and redemption;
4. proves the no-drift success path; and
[...]

PR #1463, opened 2026-08-19
fix(approvals): enforce delegated exact authority and prove fail-closed drift through the real router

ctkm-aelf:
Addresses #1457. Does **not** close it — see the AC disposition below.

Follow-up to #1460, responding to @eanz17's four-blocker audit ([comment](https://github.com/ChronoAIProject/NyxID/issues/1457#issuecomment-5314855106)).

### 1. Delegated `exact_view_digest` is mandatory (Blocker 3)

Create checked the digest only when supplied, and redeem ignored an omitted fence. #1460 even shipped a test asserting omission succeeds. That conflicts with the recorded M42 delegated contract: omitted, empty, noncanonical and mismatched must all fail closed.

Enforced by one shared validator, `ensure_delegated_exact_authority`, called from `create_request`, `observe_request` and `redeem_request` in `services/exact_service_approval_service.rs` — at the **service** boundary, not in handlers, so a future internal call site cannot skip it. `ensure_fence` stays pure.

Error precedence is deliberate: the scope gate runs before catalog resolution, but the digest requirement runs *after* exact-view membership, so delegated out-of-view targets keep their existing `NotFound` error rather than having it shadowed. Pinned by `delegated_out_of_view_error_precedence_preserved`.

Non-delegated callers are untouched; the legacy carve-out is pinned by `non_delegated_omission_still_persists_server_fence`.

### 2. Proxy-only delegated tokens rejected from exact routes (Blocker 2, related)

`catalog_delegation_service.rs:88-90` early-returns without `mcp:catalog:read`, so a delegated token carrying only `proxy` scope bypasses `validate_live_grant` entirely on this route family. It also cannot obtain a delegated catalog digest, since discovery requires that scope.

Those tokens now get `403 delegated_catalog_scope_required` on create, status and redeem. An earlier draft proposed pinning the bypass with a passing test; that was withdrawn — a green test blessing a live-grant bypass is worse than no test.

### 3. Full-router AC5 matrix (Blocker 2)

The existing coverage constructs `AuthUser` directly, so it never exercises `mw/auth.rs` — where `validate_live_grant` actually lives. A refactor could delete that call and every test would stay green.

New tests drive the real production router (`routes::build_router()`) over the real paths with real exchanged delegated JWTs, so middleware, billing extensions and `OriginalUri` are genuine. Every downstream stub has a per-route counter **plus a `.fallback()` catch-all** feeding one shared counter, so a call to any path is caught.

Matrix: success (exactly one provider call), grant absent/revoked/expired before create **and** before redeem, service deactivation, node rebinding, provider rebinding, endpoint-contract mutation, endpoint deactivation, credential revocation. Every rejection asserts HTTP status + typed error body, zero provider hits, and Mongo row state (no `ApprovalRequest` written / no `redemption` claimed).

## Deliberately not included

The escalated execution-authority work: no `execution_authority_digest`, no credential epoch, no ABA policy, no TOCTOU refactor, no contract-version bump. That is a `.v3` wire change requiring consumer coordination. Rationale in the AC5 row below.

## Rollout and compatibility

Two delegated-only breaking changes:

1. **Digest omission: 200 → 400** (`exact_view_digest_required`) at delegated create and redeem. An approved-but-unredeemed row created pre-deploy already contains the server-side digest — no data migration — but a client that omitted it must call the status endpoint, adopt the returned `exact_view_digest`, and retry. That is a real client break.
2. **Proxy-only delegated tokens: 403** (`delegated_catalog_scope_required`) where they previously passed on JWT validity alone.

No deployed or pushed consumer calls these routes with delegated tokens. The sole known consumer (M42, currently unpushed) reportedly always sends the digest and carries `mcp:catalog:read` — @eanz17's own statement, which is the best available evidence but **not proof**. Ship as an announced cutover. This is explicitly not a claim of nil blast radius; #1460's equivalent claim was already corrected on the issue.

## Blocker 1 — open coordination gate, not resolved here

`.v2` + `catalog_service_id` is the shipped producer contract. This PR does not change it. The blocker stays **open** until @eanz17 accepts that shape and evidences the M42 parser update.

Safe deploy order: NyxID `.v2` to prod first (this also unblocks the AC8/AC9 evidence the consumer is waiting on) → M42 parser updated and evidenced → M42 ships last. Aevatar-first is the only unsafe order.

Correcting #1460's blast-radius line: M42 *does* send `exact_view_digest` at create and redeem; no *deployed* client does.

If producer dual-version support is required instead, it must be end-to-end — versioned discovery **plus** create/redeem validation and persisted binding semantics. A `.v1` discovery response alone cannot redeem against a `.v2`-authoritative binding. No estimate until scoped.

## Acceptance criteria

| # | Criterion | Disposition |
|---|---|---|
| AC1 | Provider identity in projection/digest | Closed by #1460; untouched |
| AC2 | Shared resolver/projection incl. eligibility | Closed by #1460; untouched |
| AC3 | Deterministic handler snapshot | Closed by #1460; untouched |
| AC4 | Real discovery → create → redeem success | Supplemented — success path now proven through middleware |
| AC5 | Drift + live-grant fail closed before effect | **Advanced, not closed.** Enumerated classes proven through the production router with effect spies. Destination URL, credential rebind/rotation, auth/injection config, fallback nodes and `proxy_operation_policy` are outside every digest — escalated. Whether the enumerated set satisfies AC5's wording is @eanz17's call; this PR does not claim closure |
| AC6 | Omission / empty / mismatch explicit | **Delegated slice closed** — omitted, empty, noncanonical, mismatched all fail closed at create and redeem over HTTP; non-delegated carve-out pinned |
| AC7 | Discovery mutation spy | Closed by #1460; the new provider counters are AC5 evidence, not a replacement |
| AC8 | Evidence manifest / probe green | Open — post-merge human gate |
| AC9 | Release / deploy / prod `/health` | Open — human gate, out of repository scope |

## Follow-up filed separately

Adversarial review of this work surfaced a gap **not** in the original four blockers: the approval binding covers *what* executes but not *how or where*.

`exact_authority_has_drift` compares only IDs and shape digests, then `execute_tool` re-resolves the proxy target live (`mcp_service.rs:3276-3296`) and executes against current configuration. An owner or org admin can change `UserEndpoint.url`, `api_key_id`, auth method, identity/token injection or headers via `PUT /api/v1/keys/{key_id}` while an approval is pending; no digest moves, drift returns false, and the approved effect is delivered elsewhere with a different credential.

Bounds: a delegated token cannot reach the management routes, and whoever can already holds the credential. It breaks the approver's attestation rather than granting new access. Fixing it needs a producer-owned execution-authority digest, a non-secret credential epoch, an ABA policy and removal of the check-then-re-resolve race — a `.v3` contract change requiring @eanz17's coordination, which is why it is not bundled here.

## Verification

- Full backend suite against a live MongoDB replica set — counts in the comments below.
- Targeted: 4 full-router tests and 17 `exact_service_approval_service` tests pass.
- **Falsification:** neutering `ensure_delegated_exact_authority` makes the digest/scope tests fail (including through the full router, 200 vs expected 403/400); neutering the grant invalidation makes the grant matrix fail. The tests catch real regressions rather than passing vacuously.
- `cargo fmt --check` clean.

### Test robustness note

`DELEGATED_TOKEN_TTL_SECS` is a compile-time 300s while a full suite run exceeds that, so a token minted at fixture-build time could expire under parallel-execution starvation and surface as a spurious `token_expired` (2001). The fixture now re-mints token and grant at the start of each phase. Refresh always happens **before** grant mutation, never between mutation and request — otherwise the request would receive fresh authority and the assertion would prove nothing. Verified by the falsification above.


ctkm-aelf:
## Full-suite verification

Run against a live MongoDB replica set (`rs0`, transaction-capable — `test_utils::tests::transaction_test_database_supports_atomic_writes` passes, so DB-backed tests genuinely executed rather than early-returning on the missing-DB guard):

```
cargo test -p nyxid --bin nyxid-server
test result: FAILED. 5350 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 409.32s
```

The sole failure is `handlers::devices::tests::approve_handler_returns_before_slow_notification_dispatch_finishes` — pre-existing and timing-sensitive. Evidence it is unrelated: `backend/src/handlers/devices.rs` is not in this diff (`git diff --name-only main..HEAD` returns three files, none of them devices), and the test passes in isolation (`1 passed` in 7.03s). Reported rather than masked.

All `full_router_*` and `delegated_*` tests passed in that run.

### Falsification evidence

Because an earlier iteration of this work produced tests that would have passed while proving nothing, each new test was checked by deliberately breaking what it asserts:

| Mutation | Result |
|---|---|
| `ensure_delegated_exact_authority` returns `Ok(())` unconditionally | 5 tests fail, incl. through the full router (`left: 200, right: 403` and `right: 400`) with real HTTP error bodies |
| Grant invalidation removed from the grant matrix | `full_router_grant_absent_revoked_and_expired_block_create_and_redeem` fails |

Both restored afterwards and the restoration verified.

### Environment note for reviewers

`docker compose up -d` does **not** provide a usable test database here — the repo compose is a standalone `mongo:8.0`, and `connect_transaction_test_database` asserts replica-set/mongos topology and panics on standalone. Use a single-node replica set.

Separately, mongod crashed repeatedly under suite load in this environment, inside its `ftdc` (diagnostic data capture) thread — `FileNotOpen: Failed to open interim file .../diagnostic.data/...`. That produced ~100+ failures spread across untouched subsystems (`mfa_service`, `node_fanout_resolver`, `provider_service`, …), which reads exactly like a broad regression until you notice the connectivity smoke is itself failing. Starting mongod with `--setParameter diagnosticDataCollectionEnabled=false` resolved it.

PR #1470, opened 2026-08-20
feat(assistant): conversation-scoped lazy wire logs

ctkm-aelf:
## Summary

- Make assistant wire logs conversation-scoped and lazy-loaded.
- Replace the large inline response header with a short-lived Mongo-backed wire-log reference.
- Add owner-scoped fetch at GET /api/v1/assistant/wire-logs/{id}.
- Update HTTP/SSE transports and the admin panel to fetch payloads only when a row is expanded.
- Preserve bounded inline fallback behavior and keep the conversation-list route free of wire-log headers.

## Verification

- cargo fmt --all -- --check
- cargo build
- Full serial root cargo test with a dedicated Mongo replica set: backend 5,374 passed, node proxy 56 passed, CLI 1,100 passed, wizard freshness 1 passed, cloud auth 4 passed, crypto 14 passed, doc tests green.
- Focused wire-log tests: 6 passed.
- frontend npm run test: 248 files, 2,882 tests passed.
- frontend npm run lint: 0 errors, 23 existing warnings.
- frontend npm run build: tsc -b, both Vite builds, and mock-footprint assertion passed.

## Notes

- The earlier concurrent backend run hit the unrelated timing-sensitive device notification test; the definitive serial run passed it.
- The generated CLI wizard bundle was refreshed after the freshness check identified the changed frontend schema/store manifest entries.


ctkm-aelf:
Verification follow-up: implementation and local gates are complete, but the current GitHub run is blocked by findings outside this change.\n\n- Rust Clippy fails on pre-existing untouched code in `backend/src/handlers/mcp_transport.rs` (3 `result_large_err` diagnostics) and `backend/src/services/catalog_identity_service.rs` (2 diagnostics). Neither file is changed by this PR. Backend tests, format, feature builds, CLI, frontend, coverage, and wizard freshness all pass.\n- CI Pipeline is only the aggregate failure caused by `rust-clippy=failure`.\n- CodeQL reports 5 high alerts in the required generated `cli/src/wizard/assets/index.html` bundle (`Unvalidated dynamic method call` / `Insecure randomness`). The corresponding bundled code and `Math.random` occurrences are present at the same positions in the PR base bundle; this PR only refreshes the generated bundle because the changed wizard sources are in the freshness manifest.\n\nPR implementation remains conversation-scoped/lazy wire-log complete; these repository baseline findings need separate remediation or CI/CodeQL policy treatment.

PR #1479, opened 2026-08-23
fix(proxy): preserve catalog actor in delegation tokens

eanz17:
## Summary

- mint proxy delegation tokens with the canonical catalog service slug when a UserService uses a disambiguated alias
- preserve the actual UserService slug for custom and legacy services without a catalog identity
- add JWT regression coverage for both catalog-backed aliases and custom services

## Root cause

NyxID resolved the catalog-backed UserService alias correctly but signed `act.sub` with the alias slug. Downstream services validate the canonical catalog actor, so an alias such as `chrono-sandbox-aevatar` was rejected by the `chrono-sandbox` runtime with HTTP 401.

## Verification

- `cargo fmt --all -- --check`
- `cargo test -p nyxid proxy_delegation_token_uses -- --nocapture` (2 passed)
- `cargo clippy -p nyxid --all-targets -- -D warnings -A clippy::nonminimal_bool`
- `bash scripts/check-rci-backend-boundary.sh`

Local Rust 1.94 reports a pre-existing `clippy::nonminimal_bool` warning in `backend/src/services/billing/lago_client.rs`; the scoped clippy run above allowed only that unrelated baseline lint.

### DIFF
--- a/backend/src/handlers/proxy.rs
+++ b/backend/src/handlers/proxy.rs
@@ -148,6 +148,33 @@ fn auth_kind_label(method: &crate::mw::auth::AuthMethod) -> &'static str {
     }
 }
 
+/// Sign the delegation token with the downstream's canonical catalog identity.
+///
+/// UserService slugs may be disambiguated aliases, while downstream services
+/// validate `act.sub` against the catalog service they implement. Custom and
+/// legacy services have no separate catalog identity and keep their own slug.
+fn generate_proxy_delegation_token(
+    keys: &crate::crypto::jwt::JwtKeys,
+    config: &crate::config::AppConfig,
+    user_id: &uuid::Uuid,
+    scope: &str,
+    service_slug: &str,
+    catalog_service_slug: Option<&str>,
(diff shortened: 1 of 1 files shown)


chronoai-kai (approved):
Reviewed — sane and correct:

- **Right fix for a real contract break**: downstreams (e.g. `chrono-sandbox`'s `/codex/execute`) validate `act.sub` against the catalog service they implement; a disambiguated UserService alias slug in the delegation token broke that. Minting with the canonical catalog slug (falling back to the UserService slug when there is no catalog identity) matches the documented contract.
- **No authority widening**: scope, TTL (`MCP_DELEGATION_TOKEN_TTL_SECS`), and `TokenRestrictionClaims` are unchanged; only the actor identity is canonicalized. `catalog_service_slug` is resolved server-side via `load_catalog_metering_for_user_service` following `catalog_service_id` — not caller-influenced.
- **Both branches regression-tested** (`act.sub` asserted for a catalog-backed alias and for a custom service), complementing the existing resolver test that covers alias-vs-catalog slug divergence.

CI green 14/14.

PR #1484, opened 2026-08-24
fix(delegation): close exact-approval residuals and harden test isolation

eanz17:
## Summary

- Add a producer-owned `ServiceEndpoint.operation_generation`, publish it in the canonical delegated catalog v3 projection, and revalidate the same identity/generation/digest evidence during exact approval create, observe, and redeem.
- Bind all catalog-owned execution inputs in an explicit execution-authority v2 projection; consolidate live gates, claim before materialization/dispatch, and terminalize ambiguous provider outcomes without replay.
- Harden semantic effect uniqueness, provider idempotency/header handling, token-exchange caching/timeouts, and crash-safe per-test MongoDB lifecycle cleanup.

Refs #1473.

> **Draft implementation candidate only.** This PR is not release, tag, deployment, production, or mandatory machine-gate evidence. It must not close or relabel #1473.

## Behavior and security

- Catalog v3 is deterministic, typed, generic-free, and includes producer endpoint identity, generation, risk, and idempotency support.
- Exact approval request identity uses the live producer endpoint ID. Caller-supplied operation identity or generation cannot create a second semantic effect identity.
- Execution authority v2 also binds `token_exchange_config`, `service_category`, and `requires_user_credential` without persisting secret configuration literals.
- Observe and redeem share one ordered live evaluator. Redeem claims atomically before live revalidation, credential materialization, routing, or provider dispatch.
- Server-owned `Idempotency-Key` wins over defaults and caller headers when the producer contract supports it.
- Direct and node transports classify any possibly dispatched failure, stream error, missing stream terminator, hard deadline, or terminal-write ambiguity as `provider_outcome_unknown`; the approval is never automatically replayed.
- Startup backfills only missing endpoint generations. Explicit zero/negative generations remain invalid and fail closed.
- A mandatory partial unique index protects semantic effect identity; startup performs a duplicate preflight and refuses a weaker fallback index.

## Test database lifecycle

- Each test database carries a process/run lease and is dropped by a bounded worker pool after its final Mongo client is released.
- Queue overflow is retained by one deduplicating retry coordinator; transient `dropDatabase` failures receive three bounded attempts and remain registered for exit/stale recovery if all fail.
- The process pins the first writable MongoDB URI before publishing leases or cleanup targets, preventing cross-server stale sweeps and wrong-server exit cleanup.
- Startup stale sweeping uses exact managed-name parsing, quarantined canonical legacy names, owner-checked leases/drop claims, bounded work, and fail-closed claim retention after unknown drop outcomes.
[...]

PR #1487, opened 2026-08-24
feat(delegation): close #1473 residuals with rolling-safe, incremental rollout (supersedes #1484)

### LINKED ISSUE #1473: [Delegation] Residuals from #1457: operation_generation ratification, projection gaps, gate consolidation, test-DB leak
Residual items from #1457, which closed with AC1–AC7 delivered and deployed (PR #1467, merge `59172919e6b8`). These are the threads that were deliberately **not** folded into that PR, so they are not lost.

## 1. `operation_generation` has no producer-owned referent — unratified

`ExactServiceApprovalCreate.operation_generation` is caller-supplied, validated only as `>= 1`, persisted at create and echo-compared at redeem. There is no server-side value to compare against, so any drift check on it is replay self-consistency, not freshness.

PR #1467 proposes that live `endpoint_contract_digest` recomputation plus the new execution-authority digest together constitute the equivalent producer-side freshness evidence. `docs/SECURITY.md` records that this interpretation **is not ratified**.

Two ways to settle it:
- Accept the recomputation argument and mark the contract closed, or
- Bind a producer-owned revision/generation (e.g. a counter on `ServiceEndpoint`, bumped on writes) and revalidate it at redeem.

Decision needed from the acceptance holder before either path is implemented.

## 2. Execution-authority projection gaps

The projection deliberately covers exactly the enumerated closure list. Three catalog-owned inputs to execution are **not** bound, documented as known limitations in `docs/GRANULAR_APPROVALS_DESIGN.md`:

- `token_exchange_config` — read from the catalog row at proxy time; governs downstream credential minting when `auth_method == "token_exchange"`.
- `service_category` / `requires_user_credential` — gate delegated-credential resolution.

These sit outside the owner-attacker model that motivated the original fix, but the asymmetry is real: `proxy_operation_policy` and `catalog_default_headers` **are** projected, and these are not. Either bind them or record why the distinction is principled.

## 3. `execution_authority_mismatch` is misleadingly scoped

The named function covers only the **observe** path (`current_state`). `redeem_request` has a **second, independent inline gate**. A future refactor deleting the inline block on the reasonable-looking grounds that "the helper already covers it" would silently reopen the execution-authority hole with no test failure.

Mitigated for now by a doc comment and a design-doc note. The durable fix is to extract both into one named function so there is a single gate to reason about.

Related: drift caught by the observe gate leaves the approval row unclaimed, while drift caught by the redeem gate persists a `Drifted` redemption. Same logical outcome, two different persisted shapes depending on a race.

## 4. Test databases leak on killed runs (repo-level, infrastructure)

`backend/src/test_utils.rs` registers test-database cleanup that runs **only at normal process exit**, time-bounded ~3s per database. Any killed or crashed run leaks every database it created.

Observed impact during #1457 work: a dbpath accumulated **3,128 databases / 86,920 WiredTiger files**, which exhausted file handles and caused repeated `mongod` crashes with

```
"s":"F", "c":"CONTROL", "ctx":"ftdc", "msg":"Writing fatal message",
"message":"DBException::toString(): FileNotOpen: Failed to open interim file .../metrics.interim.temp"
```

Self-reinforcing: each crash leaks another run's databases, making the next crash likelier. Dropping the databases does **not** reclaim the files (WiredTiger defers until compaction) — only wiping the dbpath does.

Clean runs do drop their databases normally (~349 files after a full pass, not 30k), so the fix is crash/kill resilience, not the happy path. Options: a startup sweep of stale `nyxid_test_*` databases, or a dedicated ephemeral dbpath per run.

This one is independent of the delegation work and affects anyone running the backend suite locally.

ctkm-aelf:
## Summary

Closes the five delegation residuals of #1473 (the four listed plus the stuck-`executing` approval found in the 2026-08-25 audit) **without a maintenance window or a coordinated Aevatar release**. This supersedes draft PR #1484: it keeps that PR's engineering and replaces its rollout model.

Kept from #1484: claim-before-dispatch redemption with a 10-minute bounded effect and `provider_outcome_unknown` terminal; atomic stale-`executing` recovery (15 min); server-owned `Idempotency-Key`; execution-authority v2 (binds `token_exchange_config`, `service_category`, `requires_user_credential`); read-only observe path that never materialises credentials; producer-owned `ServiceEndpoint.operation_generation`; semantic-effect unique index; crash-safe per-test MongoDB lifecycle; the endpoint update/delete ownership fix in `handlers/endpoints.rs`.

Reworked for incremental deploy:
- **Rolling-safe authority.** New approvals store the real v1 digest in the legacy slot plus the versioned v2 binding; old replicas validate v1, new replicas v2; main-era v1-only rows are compared against the live v1 projection; pre-digest rows keep today's expiry-bounded behaviour. Rollback is safe.
- **Catalog contract stays `nyxid-delegated-operation-catalog.v2`.** `risk`, `supports_idempotency_key`, `operation_generation` are additive; discovery emits the pre-additive digest during the bounded window and both digests are accepted at create/observe/redeem (dual slots on the row so old-writer policy changes are still detected). Dated TODO for removal.
- **`operation_generation` is advisory at create** — never required, never rejected; NyxID persists the producer value and uses it for create→redeem drift only.
- **Transport classification is at-most-once safe.** Only `is_connect()`/`is_builder()` failures are `failed/provider_unreachable`; every in-flight failure (including hyper `Kind::Request` resets after the request was sent) is `provider_outcome_unknown`. Typed `McpToolExecutionOutcome::ProviderUnreachable`, no string sentinels.
- **Startup never refuses to boot** on duplicate semantic-effect groups: logs remediation, persists an `exact_service_semantic_effect_index` startup diagnostic (shown on Admin → Integrity), continues with the service-level guards.
- **Instance-spec operations stay discoverable** (`operation_generation: null`, `producer_generation_bound: false`, endpoint-contract digest remains their fence).

## Verification (independently re-run by the reviewer)

`cargo fmt --check` ✅ · `cargo clippy -D warnings` ✅ · 12 module suites re-run at `d14012fd` (exact_service_approval_service 23, delegation 37, service_endpoint_service 17, provider_token_exchange_service 29, execution_authority 7, mcp_service 114, proxy_service 125, db 31, test_utils 22, endpoints 5, approval_service 69, admin 131 = 610 tests, 0 failures) against a local rs0 replica set. CI runs the full suite on this PR.

Regression tests added for: v1 roll-forward + v1-visible drift, closed port → `provider_unreachable`, reset-after-request-sent and truncated body → `provider_outcome_unknown`, instance-spec discover/create/approve/redeem, dual-digest fence echoes, diagnostic first-detection time, sweep cooldown admission.

## Follow-up (not in scope)

`operation_catalog_digest` hashes endpoint `name`/`description`, so cosmetic edits already drift pending approvals on `main`; separating cosmetic from execution authority is a separate change.

Refs #1473. Supersedes #1484.

Closes #1473


ctkm-aelf:
Consolidated into #1494 and squash-merged into `rollup-2026-08-25-ctkm-1` as `4e31ef70` (lands on `main` via #1492). This PR's commits are included with authorship preserved; closing as superseded — the branch is left in place.

PR #1564, opened 2026-09-14
feat(catalog): add Google Drive authoring operations to the google overlay

ctkm-aelf:
## Why

Feedback on the hosted spec at `/api/v1/catalog-specs/google/openapi.json`:

> this Google openapi does not provide the ability to write and modify Google Docs, need to add it

That overlay published only 8 operations — read-only Drive, Gmail, and two Calendar ops — so nothing in it described how to create or rewrite a Google Doc. `api-google` carries no `proxy_operation_policy`, so the Drive writes already worked at the proxy; they were simply absent from the document agents read.

## What changed

**One file of spec, plus tests.** 18 operations added to `google.openapi.json`, sourced from the existing `google-drive` and `google-calendar` overlays: Drive create / update / delete / copy / export / upload / upload-content, and the Calendar surface (calendar list, calendars CRUD, event get/update/delete, freeBusy).

Google Docs authoring is described only on operations that are **new to this spec**, so no existing caller is affected:

| Operation | Note |
| --- | --- |
| `drive_create_file` | documents `application/vnd.google-apps.document` / `.spreadsheet` / `.presentation` / `.folder` |
| `drive_export_file` | points at `text/markdown` for reading a body back |
| `drive_upload_file_content` | declares a `text/markdown` body |

The last one is what makes *modify* work. `ServiceEndpoint.request_content_type` holds exactly one value and the MCP tool sends it as the request `Content-Type`; with `application/octet-stream` Drive never converts, so a Doc body could not be rewritten at all.

Round trip: `drive_create_file` → `drive_upload_file_content` → `drive_export_file` → `drive_upload_file_content`.

## Purely additive — and enforced

All **8 pre-existing operations are byte-identical**. No method, path, parameter, or body contract changed, so nothing bumps `operation_generation` and no durable exact-service approval is invalidated.

`google_overlay_keeps_its_preexisting_operations_unchanged` pins each one. Most load-bearing: `calendar_list_events` and `calendar_create_event` stay on `/calendar/v3/calendars/primary/events`. Adopting the sibling overlays' `{calendarId}` form would have added a **required path parameter** and broken every existing caller — the test comment says so, so nobody "tidies" it later.

`google-drive`, `google-calendar`, `google-gmail`, and the composed `google-workspace` specs are **untouched**. `api-google-drive`, `api-google-calendar`, `api-google-gmail`, and `api-google-workspace` keep every contract they publish today, including the binary `drive_upload_file_content` body.

## Explicitly out of scope

Structural editing (`documents.batchUpdate` — insert at index, formatting, tables). That API is served only from `docs.googleapis.com`, and a catalog service has a single `base_url`, so it needs its own seeded service. It would also need a proxy fix: `CanonicalPath::forwarding_path()` encodes `:` to `%3A`, and Google returns 400 for `documents/{id}%3AbatchUpdate` (verified against the live API, as was the `www.googleapis.com` 404 for `/v1/documents`).

## Deploy note

Overlays are `include_str!`-embedded at compile time, so the hosted spec changes only after a backend redeploy. No migration: `catalog_spec_sync` additively upserts `ServiceEndpoint` rows by operationId at startup (admin-added rows untouched), and the 18 new rows appear in `/api/v1/mcp/config` on first boot.

## Testing

- `cargo test -p nyxid catalog_spec` — 17 pass, including the three MongoDB-backed sync tests (`sync_populates_endpoints_for_seeded_services`, `additive_sync_preserves_operator_deactivation`, `spec_backed_sweep_discovers_admin_service_endpoints_once`) against the docker-compose mongod on 27018.
- Two new tests: the additive guard above, and `google_overlay_publishes_a_markdown_drive_content_write`, which fails if the body reverts to octet-stream.
- `cargo fmt` clean; clippy passed on the previous push of this branch.


ctkm-aelf:
close as no longer needed

PR #1582, opened 2026-09-16
fix(oracle): recognise the two-line GPT-6 composer pill

chronoai-kaihuei:
## The symptom

Every task on our GPT-6 workers failed in the model-selection phase. Nine workers across three machines, all with the same log:

```
model_selection reason=menu_not_opened pill_source=fallback
                pill_level=unrecognized pill_text_length=0 items=0 recognized=[]
composer_unobstructed_failed
task_failure code=browser_recovery_exhausted
probe={composer_found:true, send_found:true, pill_found:true,
       logged_in:true, url_host:"chatgpt.com", error_banner:false, turns:0}
```

The probe rules out the obvious causes. The worker is logged in, on chatgpt.com, and can see the composer, the send button and the pill.

## What the live page actually shows

I inspected a running ChatGPT tab over CDP on three worker machines. All three are identical:

```json
{
  "structural_pills_MATCHED": 1,
  "region_menu_buttons": [
    { "text": "",       "testid": "composer-plus-btn", "aria": "Add files and more" },
    { "text": "6\nPro", "testid": null,                "aria": "" }
  ]
}
```

The selector still matches. The pill now renders the **family and the level as separate text nodes**, so its `innerText` is `"6\nPro"` rather than `"GPT-5.5 Pro"`.

## Two defects compound from that

**`detectPillLevel` took the first line only.** `"6\nPro"` becomes `"6"`, the leading version is then stripped as a model number, and the label ends up empty. An ordinary GPT-6 Pro pill reports as `unrecognized`.

**`preferredModelPillIndex` then fell back to index `0`.** The composer region also holds icon-only menu buttons — `composer-plus-btn`, "Add files and more" — whose `innerText` is empty. With nothing recognised, index `0` picks that one. The worker clicks "Add files and more" instead of the pill, no model menu opens, and the run ends as `menu_not_opened` with `items=0`.

This also explains the confusing pair of values. `pill_found: true` comes from the loose probe selector at line 699 (`button.__composer-pill`), while `pill_text_length=0` is the empty label of the plus button that `preferredModelPillIndex` selected.

The `composer_unobstructed_failed` that follows is downstream of this, not a separate fault. Worth noting, because it is easy to chase the composer check instead.

## The change

`detectPillLevel` keeps first-line-only as its primary read and retries with newlines flattened **only when that yields nothing**. A menu entry whose second line is a description still matches exactly as before — that mattered, since folding descriptions in would break existing exact matches such as `Thinking`.

`preferredModelPillIndex` now prefers the first candidate that has any label at all rather than index `0`, and its numeric hint is no longer pinned to `5`.

## Verification

Against the exact values captured from the live DOM:

| | before | after |
|---|---|---|
| `detectPillLevel("6\nPro")` | `null` | `Pro` |
| `preferredModelPillIndex(["", "6\nPro"])` | `0` → the plus button | `1` → the pill |

The added tests fail on the unpatched worker (2 failures) and pass with the change. The existing 71 tests are unaffected: 71 before, 72 after.

```
git checkout -- worker.mjs && node --test worker.test.mjs   # pass 70, fail 2
node --test worker.test.mjs                                  # pass 72, fail 0
```

## Scope

Not addressed here: the same first-line truncation appears at five other call sites handling **menu item** text, where a second line is plausibly a description. Changing those needs its own evidence about what menu entries look like, so I have left them alone rather than guess.


chronoai-kai:
Landed on main via #1583 (squash commit bb93a548, released as v0.23.0): this branch was merged into channel-media-and-platform-catalog before that PR was opened, and its Oracle Worker Test passed there. Closing as merged-by-inclusion.

PR #1591, opened 2026-09-17
fix(oracle): clear the switcher marker so a worker cannot strand itself

chronoai-kaihuei:
## The symptom

One worker on `chrono-chatgpt-pro-pool` sat permanently unable to send a prompt. Its log repeated every few seconds:

```
model_selection reason=picker_unavailable pill_source=none
                pill_level=unrecognized pill_text_length=0 items=0
task ... browser failure 6/6 (operation_timeout@selecting_model)
task_failure code=browser_recovery_exhausted
```

Its two siblings on the same machine, same account, same bundle, were healthy. That ruled out the machine, the network and the account, and pointed at per-tab state.

## The cause

`readModelSwitcher` stamps `data-nyx-switcher` on whichever control it claims. On a page with no header switcher — `model-switcher-dropdown-button` was absent on every tab we inspected — that claim falls through to the composer pill itself.

Every composer-picker selector then excludes it:

```js
button.__composer-pill[aria-haspopup="menu"]:not([data-nyx-switcher])
```

So `pickerSnapshot` finds no candidates and reports `pill_source=none`. Measured on the stuck tab:

```
pills ignoring the marker : 1   (class "__composer-pill", text "6\nPro")
pills as the code selects : 0
elements carrying marker  : 1
```

The worker cannot recover on its own. `readModelSwitcher` is the only place that clears markers, and it re-stamps the same pill on the next attempt, so every retry rebuilds the condition instead of breaking it.

## The fix

The marker is genuinely needed while the switcher interaction runs — `selectModelSwitcher` locates its trigger through it. It just must not outlive the call. This clears it in the cleanup that already runs in `finally`, next to the menu dismissal.

## Verification

Live, on the stuck worker, removing the attribute alone:

```
BEFORE  pills_as_code_selects=0   elements_carrying_marker=1
AFTER   markers_cleared=1   pills_now_selectable=1   pill_text="6\nPro"
```

The tab recovered with no restart, and the pool answered the next task on the first attempt.

`node --test worker.test.mjs` → **76 pass, 0 fail**.

One caveat worth stating plainly: the tests confirm no regression, they do not exercise this path. The marker logic runs inside `page.evaluate` browser context, which the current suite has no way to reach. The evidence here is DOM observation on a live worker, not a unit test.

## Known limit

A worker killed mid-interaction still leaves the attribute behind, because no `finally` runs. Self-healing inside `pickerSnapshot` would cover that case too, but it would blur the header-versus-composer distinction the marker exists to draw. Deliberately not done here.

## Related

Follows #1582, which fixed the two-line `"6\nPro"` pill label. Same worker, different failure — that one misread the pill, this one cannot see it at all.


chronoai-kaihuei:
Reopening: the bug recurred today on the 0.24.1 bundle, and this branch still applies cleanly.

**It is still present on `main`.** `data-nyx-switcher` has exactly one clear site, inside `readModelSwitcher`:

```
1484  body.querySelectorAll('[data-nyx-switcher]').forEach(el => el.removeAttribute('data-nyx-switcher'))   ← only clear
1500  if (trigger) trigger.setAttribute('data-nyx-switcher', '')                                            ← stamp
1343  PILL_SELECTOR      = 'button.__composer-pill[...]:not([data-nyx-switcher])'                           ← excluded
1734  pills              = '...__composer-pill[...]:not([data-nyx-switcher])'                               ← excluded
1735  candidates         = 'button[aria-haspopup="menu"]:not([data-nyx-switcher])'                          ← excluded
1787  region.locator('button[aria-haspopup="menu"]:visible:not([data-nyx-switcher])')                       ← excluded
```

Nothing clears the attribute when `selectModelSwitcher` returns, so a run that ends anywhere other than through `readModelSwitcher` leaves it stamped. On a page with no header switcher the stamp lands on the composer pill, every selector above then skips it, and `pickerSnapshot` reports `pill_source=none`. The worker retries, re-stamps, and rebuilds the same state — it cannot recover on its own.

**Recurrence, 2026-09-18.** Two of twelve workers in the same pool, both on `0.24.1+15e753a955dd`, both `online / login=yes / chrome=yes`:

```
heca-01a08ac9-1       last_error = operation_timeout@selecting_model
heca-01a08ac9-2297…   last_error = operation_timeout@selecting_model
heca-01a08ac9-1091…   last_error = -      ← same machine, healthy
```

One healthy sibling on the same machine and account again rules out the host, the network and the session, which is what pointed at per-tab DOM state the first time.

**Prior measurement**, from the original report — a stuck tab, before and after removing the attribute and nothing else:

```
BEFORE  pills ignoring the marker = 1   ("__composer-pill", text "6\nPro")
        pills as the code selects = 0
        elements carrying marker  = 1
AFTER   markers cleared = 1   pills now selectable = 1
```

The tab recovered with no restart and answered the next task on the first attempt. Two more tabs on two other machines showed the identical signature the same day and were fixed the same way.

The operational cost is that each occurrence strands a paid Pro seat until someone clears it by hand or restarts its Chrome. We have now done that on four tabs across three machines in two days.

I could not find a review comment explaining the close, so I may well have missed a reason — if the approach is wrong, I am glad to take direction. Two alternatives, if the concern was the placement:

1. Clear inside `pickerSnapshot` when the strict selector finds nothing but the marker-ignoring one finds a pill. This also covers a worker killed mid-interaction, which the `finally` cannot. I left it out because it blurs the header-versus-composer distinction the marker exists to draw.
2. Scope the marker to the handle rather than the DOM, so no attribute outlives the call at all. Bigger change, but the failure mode stops being possible.

Happy to switch to either.

chronoai-kaihuei:
CI caught a real defect in this branch. I've rewritten it — force-pushed, now rebased on `main` (d104ce8).

**What was wrong.** The first revision cleared `data-nyx-switcher` unconditionally in `selectModelSwitcher`'s `finally`. That broke five `session.browser.test.mjs` tests in the *compact model switcher* group:

```
✖ strict worker delivery reads the five-button composer without inferring Extended
✖ strict worker delivery selects Pro from Thinking without inferring effort
✖ selects Pro from Thinking and verifies a real fallback effort pill
✖ pre-send family change fails before the durable Send fence     expected gpt_5_5_pro, got gpt_6_pro
✖ pre-send absent change fails before the durable Send fence     expected absent,      got gpt_6_pro
```

I reproduced it locally and bisected it against the same commit:

```
main (d104ce8)                       compact model switcher  17 pass  0 fail
main + the old unconditional clear   compact model switcher  12 pass  5 fail
```

So the reviewer's instinct to close this was right about the patch, even if the underlying bug is real. The marker has to **survive** `selectModelSwitcher` — the effort-pill step and the pre-send verification read both depend on it to avoid re-picking the switcher they just used. I had only run `worker.test.mjs` and never the browser suite, which is why I missed it. My apologies for the noise.

**What the branch does now.** The recovery moved into `pickerSnapshot`, gated on the one state that cannot arise legitimately — no unmarked pill anywhere, but a marked composer pill present:

```js
if (!pills.length) {
  const stranded = [...body.querySelectorAll(
    'button.__composer-pill[aria-haspopup="menu"][data-nyx-switcher]')].filter(visible);
  if (stranded.length) {
    stranded.forEach((el) => el.removeAttribute('data-nyx-switcher'));
    pills = stranded;
  }
}
```

Whenever an unmarked candidate exists the marker is untouched, so the lifetime the tests encode is preserved. It also covers the case the `finally` never could: a worker killed mid-interaction, where no cleanup runs at all.

**Verification**

```
worker.test.mjs           77 pass   0 fail
session.browser.test.mjs  104 pass   0 fail
```

The underlying fault and its field evidence are unchanged from the earlier comment: four tabs stranded across three machines, on 0.23.1 and again on 0.24.1, each recovered by removing the attribute alone.

chronoai-kaihuei:
Closing: this branch is now empty. The self-heal it proposed was taken into `main` verbatim by release commit 0fa8299 ("prepare 0.25.2 with X polling and oracle fixes", #1607), so there is no diff left against `main`.

Field note for anyone following the thread. The self-heal is working — the stranded tab that used to log `pill_source=none` now logs `pill_source=structural` and reads the pill correctly. But those workers still failed, on a second and unrelated bug: `pillShowsLevel` read only the first line of the label, so a two-line `"6\nPro"` pill reported that it was not showing Pro even though `detectPillLevel` said it was. That is fixed in #1619, with a regression test that fails without it.

PR #1619, opened 2026-09-18
fix(oracle): restore model selection against the current ChatGPT composer

chronoai-kaihuei:
## The symptom

Three workers on one machine failed every task. Each attempt logged:

```
model_selection reason=level_unavailable pill_source=structural
                pill_level=Pro pill_text_length=5 items=5 recognized=[Pro]
```

That line is self-contradictory. The picker **found** the pill, classified it as **Pro**, opened a menu of five entries, and **recognised Pro among them** — then reported the level unavailable.

Deleting a worker and installing a fresh one, with a fresh Chrome profile and a fresh sign-in, reproduced it exactly. That ruled out profile, session and account. The accounts are genuine `planType: pro`.

## The cause

Two functions that must agree, disagree. `detectPillLevel` understands a pill that renders family and level on separate lines (`"6\nPro"`). `pillShowsLevel` took only the first line:

```js
pillText = String(pillText || "").trim().split(/\r?\n/)[0];
```

which leaves `"6"` — classified as `null`. So a pill already showing Pro reported that it wasn't:

| pill text | `detectPillLevel` | `pillShowsLevel` |
|---|---|---|
| `"6\nPro"` | `Pro` | **`false`** |
| `"6 Pro"` | `Pro` | `true` |
| `"Pro"` | `Pro` | `true` |
| `"ChatGPT 6 Pro"` | `Pro` | `true` |

Two consequences, both fatal:

`selectModelInner` never reaches its `already_selected` path, so it opens the picker and hunts for a `Pro Extended` entry that this account's menu does not offer → `clicked = false` → `level_unavailable`.

`result.verified` also calls `pillShowsLevel`, so the selection could never verify even if it had landed.

This only bites where the composer pill wraps onto two lines, which is why twelve workers on other machines were unaffected while all three on this one failed.

## The fix

Keep the whole label. `detectPillLevel` already handles both shapes; matching it here keeps the pair in step. The Standard/Extended split still discriminates across the newline:

```
"6\nPro Standard"  shows(pro)=false  shows(standard)=true
"6\nPro Extended"  shows(pro)=true   shows(standard)=false
"6\nInstant"       shows(pro)=false
```

## Relationship to #1582

This is the half of #1582 that was missed. That PR taught `detectPillLevel` to join the lines and left `pillShowsLevel` truncating them. The two have been out of step since.

## Verification

The regression test fails on the old line-one behaviour and passes on the new — checked both ways, not just assumed:

```
with the bug restored:  tests 1  pass 0  fail 1
with the fix:           tests 1  pass 1  fail 0
```

Full suites:

```
worker.test.mjs            78 pass  0 fail
session.browser.test.mjs  104 pass  0 fail
```


chronoai-kaihuei:
**Rewritten — my first version was wrong, and this repo'"'"'s own tests said so.**

I had loosened `pillShowsLevel` to stop truncating multi-line labels. Then I tried the same loosening on `effortMetadata` and it broke two existing tests:

```
✖ compact numeric labels require adaptation for model matching and never imply effort
✖ compact model labels accept only whole numeric family and known tier labels
     + actual 'pro'   - expected 'unrecognized'
```

Those tests are explicit:

```js
assert.equal(effortMetadata('6\nPro'),    'unrecognized');
assert.equal(effortMetadata('GPT 6 Pro'), 'pro');
```

Refusing an un-adapted compact label is **deliberate**, not an oversight. Loosening the helpers would have broken a contract this repo states plainly, so I reverted it. Apologies — I should have read those tests before proposing the first version.

**The actual gap.** `readModelSwitcher` (line 1500) and `chooseSwitcherEntry` (line 1462) both run a compact label through `compactModelLabel` before anything reads it. `pickerSnapshot` returns the composer pill'"'"'s raw `innerText`. That one read site skips the adaptation the other two perform:

```
pillShowsLevel("6\nPro",    targets) = false    effortMetadata("6\nPro")    = unrecognized
pillShowsLevel("GPT 6 Pro", targets) = true     effortMetadata("GPT 6 Pro") = pro
```

So the fix belongs at the read site, and it is one expression — `compactModelLabel(raw) || raw` — matching what the other two do. No helper contract changes; a raw compact label is still refused everywhere else.

This also explains the second symptom I could not account for before: `Observed effort: unrecognized` on every worker. Same cause, same fix.

**Field impact.** A 15-worker pool is fully down on this — every task fails `model_unavailable@selecting_model` after being rerouted through all 15:

```
Attempts: 15 (infrastructure retries 0/3)   Capacity reroutes: 14
Observed switcher: gpt_6_pro   Observed effort: unrecognized
```

```
worker.test.mjs            77 pass  0 fail
session.browser.test.mjs  104 pass  0 fail
```

One gap I will not paper over: the adaptation itself has no direct test, because `pickerSnapshot` only runs against a live page. The browser fixtures use single-line labels, so none of them exercise a compact two-line pill. I am happy to add a fixture for it if you would like that in this PR rather than a follow-up.

chronoai-kaihuei:
**Confirmed against a live worker**, not just a fixture. Three workers on one machine, all identical:

```
headerSwitcher: false
one __composer-pill[aria-haspopup="menu"]   text = "6\nPro"   marked = true
```

Evaluated in the page itself, using the page's own installed helper:

```json
{"raw":"6\nPro","adapted":"GPT 6 Pro","helperPresent":true,
 "markedOnPill":true,"currentSelectorFinds":0}
```

So `window.__nyx.compactModelLabel` — already present on the page — turns the live pill text into exactly the form the helpers accept. This PR just calls it at the one read site that doesn't. Feeding the live text through the helpers:

```
raw      "6\nPro"      pillShowsLevel = false   effortMetadata = unrecognized
adapted  "GPT 6 Pro"   pillShowsLevel = true    effortMetadata = pro
```

Both field symptoms — `level_unavailable` and `Observed effort: unrecognized` — come from that one line, and both clear with it.

Worth noting `currentSelectorFinds: 0`: the marker is stranded on the pill, so the strict selector matches nothing. The self-heal from #1607 recovers the pill; this PR makes what it recovers actually verify. The two compose — neither alone restores the worker.

For context on urgency: the pool this came from is fully down, all 15 workers, every task failing `model_unavailable@selecting_model` after being rerouted through every one of them.

chronoai-kaihuei:
**Added a second commit** — a different fault, found in the same pool today, with the same shape: a worker that cannot select a model and cannot recover on its own.

`clearComposerDraft` already runs before model selection and its comment states the reason exactly: a long leftover draft makes selection exceed `MODEL_SELECT_TIMEOUT_MS` and the task dies `operation_timeout@selecting_model`. It just doesn't hold at the sizes that actually occur.

The clear ends in `fill("")`, which drives a select-all and delete through React. On a composer holding tens of thousands of characters that doesn't finish inside `PRE_SEND_ACTION_MS`, and the helper is deliberately best-effort — so the timeout is swallowed and the draft stays exactly where it was. Selection then runs against the heavy composer and times out. The guard fails at precisely the size it exists for.

**It is worse than one stuck worker.** The draft outlives a browser relaunch, so recovery reinstates it, and the task moves on to strand the next worker.

Measured today. One 83,046-character prompt walked a 15-worker pool:

```
task 46ea4587 paused for browser recovery (operation_timeout@selecting_model)
stopping 9 Chrome process(es) for a real relaunch
model_selection reason=level_unavailable ...        ← draft still there after relaunch
```

I cleared the tab by hand — `draft 83046 → 0`, `selectable 0 → 1` — and that worker recovered. The task then moved to the next worker and stranded it with the identical 83,046-character draft. It only stopped when I cancelled the task.

The fix reads the draft's length alongside its head and, above `COMPOSER_FAST_CLEAR_CHARS`, empties it with one DOM assignment plus an `input` event rather than typing it away. Ordinary drafts keep the existing path, so interactive behaviour is unchanged where it already worked. The prompt is re-typed after selection, so nothing real is dropped either way.

```
worker.test.mjs            78 pass  0 fail
session.browser.test.mjs  104 pass  0 fail
```

Both commits are on this branch now, as requested by the operator running the affected fleet:

```
8247523  drop an oversized leftover draft in one assignment
ebc5443  adapt the compact composer pill label before reading it
```

Being straight about coverage: the threshold predicate is unit-tested, but the fast-clear path itself only runs against a live composer, so no fixture exercises it — same gap I flagged on the first commit. Happy to add browser fixtures for both if you'd rather they land together.

chronoai-kaihuei:
**Third commit added — and this is the one that actually restores service.** All three are verified end to end on a live worker.

ChatGPT has removed the Pro Standard / Pro Extended split on some accounts. Captured from a live Pro account today, the composer pill's menu is now:

```
"6 Pro"                          aria-checked unset   ← heading, not clickable
""                                                    ← separator
"Latest"                         aria-checked true
"GPT-5.6 Sol"                    aria-checked false
"GPT-5.5 Leaving on October 14"  aria-checked false
```

`modelLevelTargets` still asks for `Pro Extended` on every Pro request — no pool setting yields a bare `Pro` — so `clickMatchingLevel` can never succeed against this menu. The worker reported `level_unavailable` while its pill sat on the requested model the whole time. The `recognized=[Pro]` that made those logs look self-contradictory came from that unclickable heading.

When nothing in the menu offers the split, the pill already shows the requested level and there is nothing left to click: close the menu, report `already_selected`. Where the split does exist the hunt is untouched — `splitTierOffered` separates the two, and it reads the localized labels (`Pro 扩展`, `Pro 标准`) too.

**Before and after on the same worker, same page, same menu:**

```
before   model_selection reason=level_unavailable  pill_level=Pro items=5 recognized=[Pro]
after    model_selection reason=already_selected selected=Pro items=5 recognized=[Pro]
         prompt 3ce8f331 -> saved (3 chars)
```

The three commits compose, and the order matters — without the label adaptation the observed pill is `"6\nPro"`, `pillShowsLevel` is false, and this branch is never reached. The progression is visible in one worker's log across the session: `pill_text_length=5` → `9` → `reason=already_selected`.

```
103f3cf  accept a Pro pill when the menu offers no split tier
8247523  drop an oversized leftover draft in one assignment
ebc5443  adapt the compact composer pill label before reading it

worker.test.mjs            79 pass  0 fail
session.browser.test.mjs  104 pass  0 fail
```

**Operational context for whoever reviews this.** A 15-worker pool was fully down on the third bug — every task rerouted through all 15 workers and failed — and is currently running with `require_model_match=false` as a stopgap. These three commits are what let verification be turned back on.

Coverage, stated plainly: the helpers (`splitTierOffered`, `draftNeedsFastClear`) are unit-tested including the live menu shape above, but the three code paths that consume them only run against a real composer, so no fixture exercises them. The evidence for those is the live before/after. I'm glad to add browser fixtures if you'd prefer them in this PR.

chronoai-kai:
Landed on main via #1629 (squash commit d3573032, released as v0.29.0); closing as merged through that PR.

PR #1628, opened 2026-09-21
fix(mcp): let agent API keys read /mcp/config and the catalog

chronoai-kai:
## Summary

A user reported that agent API keys could not see the OpenAPI spec for custom endpoints or user-overridden specs. The root cause was routing: `GET /api/v1/mcp/config` was nested inside the human-only router, which layers `reject_api_key_tokens`, even though the handler already derives API-key node and service scope and `docs/API.md` promises parity with stateless MCP `tools/list`. The `/keys` half of the report was stale; those inventory GETs already accept agent keys.

- Move `GET /mcp/config` into the shared router beside the inventory reads. Sessions, proxy-scoped access tokens, general and platform API keys, and non-Curation service accounts are accepted. Scheduled-invocation keys, Curation service accounts, delegated tokens, and relay tokens remain rejected. Chat keys use the same discovery policy as `tools/list`, and a dangling chat-key lookup maps to 401 like the MCP transport.
- Move `GET /catalog`, `/catalog/{slug}`, and `/catalog/{slug}/endpoints` into the shared router so `nyxid catalog list|show|endpoints` works from an agent-key profile. No proxy scope is required. Service accounts stay rejected because catalog detail resolves restricted platform grants through a User actor. Catalog writes stay human-only.
- Add `catalog_discovery_service` so API-key callers resolve instance-backed private catalog access and mounted-spec fallback only through their effective service allowlist (including auto-connected expansion) and Member/Admin org scopes. Org-owned keys act as the org. Catalog reads perform no provisioning or pending-OAuth reconciliation, proven with MongoDB command monitoring.
- Align MCP `nyx__discover_services` with the REST catalog visibility rule for every caller. This closes a pre-existing exposure where private template metadata was listed to any authenticated caller. Restricted keys suppress connected templates only for instances in their visible inventory; unrestricted keys receive the same result as the owner's session.
- Relocate the existing catalog handler tests unchanged and add sixteen router-level regressions across two new test modules. Update `docs/API.md`, `docs/AGENT_ISOLATION.md`, `docs/API_DISCOVERY.md`, and `CLAUDE.md`.

## Test Plan

- [x] `cargo fmt --all -- --check` and `cargo clippy -p nyxid --all-targets -- -D warnings`: clean.
- [x] New modules `config_routes_tests` (8) and `catalog_routes_tests` (8), plus `handlers::mcp`, `handlers::catalog`, `catalog_service`, `mcp_service`, `routes`, `mw::auth`: all pass on a single-node replica set matching CI.
- [x] Mutation check: reverting only the route move makes the new `/mcp/config` test fail with the reported 403.
- [x] Full backend package: 6413 passed, 0 failed, 1 existing ignored benchmark.
- [x] `cargo test -p nyxid-cli catalog`: 42 passed.
- [ ] GitHub CI on this head.

## Checklist

- [x] Code follows the project's architecture rules.
- [x] No production credentials included.
- [x] Product documentation updated.


chronoai-kai:
Landed on main via #1629 (squash commit d3573032, released as v0.29.0); closing as merged through that PR.

